PR #2, opened 2021-11-23
Adding possibility of setting custom name to the patched service

Gmerold:
Magma code seems to be coupled with predefined K8s services names. By default, Juju sets service name to application (charm) name. Since charms naming convention in charmed-magma forces charms names different from services names defined by Magma, there's a need for a mechanism which would allow setting custom name for the patched service.
This PR implements such mechanism.


jnsgruk (approved):
I was initially unsure about this, but I think so long as the default is still to use the app name then we're fine. Hoping this library's purpose will be obviated by the Juju scriptlets/templates work soon anyway!

PR #3, opened 2022-01-18
Adds option to specify labels and selectors to service patches

gruyaume:
- Adds option to add new labels and selector when patching a service.
- Fixes an issue where when service_names were provided, the service wasn't patched correctly

### DIFF
--- a/lib/charms/observability_libs/v0/kubernetes_service_patch.py
+++ b/lib/charms/observability_libs/v0/kubernetes_service_patch.py
@@ -231,8 +231,8 @@ class KubernetesServicePatch(Object):
 
     def _delete_and_create_service(self, client: Client):
         service = client.get(Service, self._app, namespace=self._namespace)
-        service.metadata.name = self.service_name
-        service.metadata.resourceVersion = service.metadata.uid = None
+        service.metadata.name = self.service_name  # type: ignore[attr-defined]
+        service.metadata.resourceVersion = service.metadata.uid = None  # type: ignore[attr-defined]   # noqa: E501
         client.delete(Service, self._app, namespace=self._namespace)
         client.create(service)
 
--- a/lib/charms/observability_libs/v0/kubernetes_service_patch.py
+++ b/lib/charms/observability_libs/v0/kubernetes_service_patch.py
@@ -12,7 +12,7 @@ When modifying the default set of resources managed by Juju, one must consider t
 charm. In this case, any modifications to the default service (created during deployment), will
 be overwritten during a charm upgrade.
 
-When intialised, this library binds a handler to the parent charm's `install` and `upgrade_charm`
+When initialised, this library binds a handler to the parent charm's `install` and `upgrade_charm`
 events which applies the patch to the cluster. This should ensure that the service ports are
 correct throughout the charm's life.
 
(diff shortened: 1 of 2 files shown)


lib/charms/observability_libs/v0/kubernetes_service_patch.py
```diff
@@ -157,22 +168,32 @@ def _service_object(
...
+                "app.kubernetes.io/name" is set to the service name)
+            additional_selectors: Selectors to be added to the kubernetes service (by default only
+                "app.kubernetes.io/name" is set to the service name)
 
         Returns:
             Service: A valid representation of a Kubernetes Service with the correct ports.
         """
         if not service_name:
             service_name = self._app
+        labels = {"app.kubernetes.io/name": service_name}
```
rbarry82, line 181: This is not correct, and is potentially bad.

A merge patch is _probably_ ok, but the risk of `_delete_and_create_service` clobbering this is bad.

We should fetch the actual labels and selectors before blindly trying to update them. Juju **does** set values for these, and they may change out from underneath us. Even the default values include more than `app.kubernetes.io/name: foo`

```yaml
apiVersion: v1
kind: Service
metadata:
  labels:
    app.juju.is/created-by: prometheus
    app.kubernetes.io/managed-by: juju
    app.kubernetes.io/name: prometheus
  selector:
    app.kubernetes.io/name: prometheus
```

We'd have to instantiate the `Client` object first and actually query it, but the safety is worthwhile

gruyaume: In the case where the user provides a `service_name`, the `_delete_and_create_service` first `gets` a service before deleting it, then recreates the same service with a different name. This new service will then those same juju specific labels/selectors as before. Finally the service is merged patched with the new user provided labels.

rbarry82: That's reassuring. I didn't look into the exact return from `client.Get`

lib/charms/observability_libs/v0/kubernetes_service_patch.py
```diff
@@ -197,7 +218,9 @@ def _patch(self, _) -> None:
 
         client = Client()
         try:
-            client.patch(Service, self._app, self.service, patch_type=PatchType.MERGE)
+            if self.service_name != self._app:
+                self._delete_and_create_service(client)
```
jnsgruk, line 222: It looks like you're missing an else here? Presumably no need to patch the service you've just created?

lib/charms/observability_libs/v0/kubernetes_service_patch.py
```diff
@@ -221,6 +251,10 @@ def is_patched(self) -> bool:
         fetched_ports = [(p.port, p.targetPort) for p in service.spec.ports]  # type: ignore[attr-defined]  # noqa: E501
         return expected_ports == fetched_ports
 
+    def _on_remove(self, event):
```
jnsgruk, line 254: Curious why you added this? Assuming the juju labels persist, juju should automatically remove the service when the charm is removed?

lib/charms/observability_libs/v0/kubernetes_service_patch.py
gruyaume: Well it wasn't really a mistake. The service we create during `_delete_and_create_service` doesn't have the ports/labels that the user provided. It's simply the same old service with a new name.

lib/charms/observability_libs/v0/kubernetes_service_patch.py
gruyaume: Labels persist indeed but juju doesn't automatically remove the service on remove. I believe that it's because the 'app.kubernetes.io/name' label points to the new service name and that juju looks for the application that it's trying to remove.

rbarry82 (approved):

jnsgruk (approved):
LGTM, thanks @gruyaume :)

PR #4, opened 2022-02-10
Add ability to optionally specify protocols for service ports

stonepreston:
Currently one is unable to create anything other than TCP service ports. This PR adds the ability to optionally specify a list of protocols (TCP, UDP, or SCTP) to go along with the list of service ports. The list of protocols are provided as a keyword argument. I would have preferred to provide the protocol as part of the tuple that gets passed in for each port, but couldn't think of a way to easily make it optional with the logic that is currently in place:
```python
ServicePort(
    name=p[0],
    port=p[1],
    targetPort=p[2] if len(p) > 2 else p[1],  # type: ignore[misc]
    nodePort=p[3] if len(p) > 3 else None,  # type: ignore[arg-type, misc]
)
for p in ports
```
If the protocol was made mandatory, then this wouldn't be an issue as the protocol could be added to the beginning of the port tuple. But since I wanted to avoid breaking the current API I opted to go the keyword argument  route.

### DIFF
--- a/lib/charms/observability_libs/v0/kubernetes_service_patch.py
+++ b/lib/charms/observability_libs/v0/kubernetes_service_patch.py
@@ -16,14 +16,12 @@ When initialised, this library binds a handler to the parent charm's `install` a
 events which applies the patch to the cluster. This should ensure that the service ports are
 correct throughout the charm's life.
 
-The constructor simply takes a reference to the parent charm, and a list of tuples that each define
-a port for the service, where each tuple contains:
+The constructor simply takes a reference to the parent charm, and a list of Lightkube ServicePorts
+that each define a port for the service. For information regarding the Lightkube ServicePort model,
+please visit https://gtsystem.github.io/lightkube-models/1.23/models/core_v1/#serviceport.
 
-- a name for the port
-- port for the service to listen on
-- optionally: a targetPort for the service (the port in the container!)
-- optionally: a nodePort for the service (for NodePort or LoadBalancer services only!)
-- optionally: a name of the service (in case service name needs to be patched as well)
+Optionally, a name of the service (in case service name needs to be patched as well), labels,
+selectors, and annotations can be provided as keyword arguments.
 
 ## Getting Started
 
@@ -45,11 +43,13 @@ For ClusterIP services:
 ```python
 # ...
 from charms.observability_libs.v0.kubernetes_service_patch import KubernetesServicePatch
+from lightkube.models.core_v1 import ServicePort
 
 class SomeCharm(CharmBase):
   def __init__(self, *args):
     # ...
-    self.service_patcher = KubernetesServicePatch(self, [(f"{self.app.name}", 8080)])
+    port = ServicePort(443, name=f"{self.app.name}")
+    self.service_patcher = KubernetesServicePatch(self, [port])
     # ...
 ```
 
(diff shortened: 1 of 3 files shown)


jnsgruk (dismissed):
Looks good to me, though if possible it would be good to see an example in the big old docstring up front :)

lib/charms/observability_libs/v0/kubernetes_service_patch.py
```diff
@@ -67,6 +70,25 @@ def __init__(self, *args):
...
+# ...
+from charms.observability_libs.v0.kubernetes_service_patch import KubernetesServicePatch
+
+class SomeCharm(CharmBase):
+  def __init__(self, *args):
+    # ...
+    self.service_patcher = KubernetesServicePatch(
+        self,
+        [("svc1", 123, 123), ("svc2", 124, 124), ("svc3", 125, 125), ("svc4", 126, 126)],
+        protocols=["TCP", "UDP", "SCTP", "TCP"],
```
jnsgruk, line 87: Hmm, I'm a little uncomfortable with this - it feels a little gross to have to match the placement of the defined ports with this array.

I think this lib has grown specifically that we *could* consider moving to a new `v1` API version and break the old? @gruyaume any thoughts? I'd be interested to see what that might look like. Perhaps defining the services as a dict rather than a tuple?

@simskij this would require a change to bunch of your charms, although not *much* of a change, and only when the developers choose to bring the new library version in.

mmanciop: We could NOT have a separate array for the ports, but accept ports in an extended format as `<protocol>/<port>`. Ports specified only as an int (or string that converts to an int) are assumed to be http for backwards compat.

mmanciop: Fundamentally we have PortDefinition already specified for us to extend as we like. Adding the protocol there, either as part of the port value of (maybe better) as another element of the tuple, sounds the best option to me.

About PortDefinition: high time it becomes a DataClass or similar. Having it defined as a Union of tuples is pretty gross :D

jnsgruk: Okay so supporting `"80/TCP"` and `80` seems fine to me - we can just assume TCP as the default

jnsgruk: And agree on PortDefinition - when I originally implemented this it was much simpler!

mmanciop: But then only in the source port? What would happen is source and target port have to different protocols? I gave it some more thought and I am thinking that having a separate protocol field in PortDefinition might be cleaner and less surprising (that’s how K8s does it after all).

gruyaume: I'd agree with @mmanciop to create a dataclass for ports. Something like this would be pretty clean.


```
@dataclass
class Port:
    name: str
    port: int
    target_port: int = None
    node_port: int = None
    protocol: str = "TCP"
```

jnsgruk: Or perhaps at this point we just use the ServicePort from lightkube itself?! https://gtsystem.github.io/lightkube-models/1.17/models/core_v1/#serviceport

jnsgruk: @stonepreston up for giving it a shot?

stonepreston: sure thing. Should be able to get the PR updated in the next few hours. I like the idea of just using the service port from lightkube, but if you can think of a reason to create a custom class I can do that as well.

jnsgruk: Seems like it would work pretty nicely to me -- @gruyaume/@mmanciop?

gruyaume: Sounds good to me!

stonepreston: Newest commit has the ServicePort changes

jnsgruk:
Merged and published, thanks @stonepreston!

https://charmhub.io/observability-libs/libraries/kubernetes_service_patch

PR #9, opened 2022-03-23
Add Service Discovery Event

manadart:
## Issue
We need a mechanism to dynamically discover Prometheus metrics endpoints for applications deployed to a k8s cluster.

Rather than co-opt an existing Juju hook to handle such discovery events, it is desirable to have an event dedicated to this purpose.

## Solution
Here we use the Ops facility for custom events to add a new event and event source extension, so that charms can use it to handle metrics endpoint changes.

## Context
From the charm's tools directory, such an event can be dispatched by using the `juju-run` command like this:
```bash
./juju-run -u my-application/0 "JUJU_DISPATCH_PATH=hooks/metrics_endpoint_change ./dispatch"
```

## Testing Instructions
See the usage instructions added with the new lib. Follow them to:
- Instantiate `MetricsEndpointObserver` in the initialisation logic for an existing charm.
- Add a handler for the event via the observer's `on` member.
- Deploy a unit of the charm.
- SSH to the unit.
- For the unit's tools directory, dispatch the event as described in _Context_ above.
- Verify that the handler was invoked.

## Release Notes
Adds a new custom hook event that can be recruited to handle upcoming metrics endpoint changes within a K8s cluster.

### DIFF
--- a/lib/charms/observability_libs/v0/metrics_endpoint_discovery.py
+++ b/lib/charms/observability_libs/v0/metrics_endpoint_discovery.py
@@ -1,26 +1,40 @@
 # Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-"""TODO: Add a proper docstring here.
+"""# MetricsEndpointDiscovery Library.
 
-This is a placeholder docstring for this charm library. Docstrings are
-presented on Charmhub and updated whenever you push a new version of the
-library.
+This library provides functionality for discovering metrics endpoints exposed
+by applications deployed to a Kubernetes cluster.
 
-Complete documentation about creating and documenting libraries can be found
-in the SDK docs at https://juju.is/docs/sdk/libraries.
+It comprises:
+- A custom event and event source for handling metrics endpoint changes.
+- Logic to observe cluster events and emit the events as appropriate.
 
-See `charmcraft publish-lib` and `charmcraft fetch-lib` for details of how to
-share and consume charm libraries. They serve to enhance collaboration
-between charmers. Use a charmer's libraries for classes that handle
-integration with their charm.
+## Using the Library
 
-Bear in mind that new revisions of the different major API versions (v0, v1,
-v2 etc) are maintained independently.  You can continue to update v0 and v1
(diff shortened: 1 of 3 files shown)


simskij:
Testing instructions seems to contain some copy/paste artifacts. :sweat_smile: To get it to work I had to replace `hooks/discovery` with `hooks/metrics_endpoint_change`.

Is there a way we could change this to not have to extend `self.on` manually, for instance through a `MetricsEndpointDiscovery` class we could subscribe to in a fashion similar to [what is being done here](https://github.com/canonical/prometheus-k8s-operator/blob/8b8e23b929e27d6ffc909c79dc480cb39726c760/src/charm.py#L90)?

lib/charms/observability_libs/v0/metrics_endpoint_discovery.py
```diff
@@ -1,27 +1,41 @@
 # Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-"""TODO: Add a proper docstring here.
+"""# MetricsEndpointDiscovery Library.
```
simskij, line 4: Testing instructions seems to contain some copy/paste artifacts. sweat_smile To get it to work I had to replace hooks/discovery with hooks/metrics_endpoint_change. 😅 

Is there a way we could change this to not have to extend self.on manually, for instance through a MetricsEndpointDiscovery class we could subscribe to in a fashion similar to [what is being done here](https://github.com/canonical/prometheus-k8s-operator/blob/8b8e23b929e27d6ffc909c79dc480cb39726c760/src/charm.py#L90)?

simskij (changes requested):

lib/charms/observability_libs/v0/metrics_endpoint_discovery.py
mmanciop: I too think that this would benefit from the constructor-centric approach we use in the other O11y libs.

manadart: I believe I've addressed everything.

PR #11, opened 2022-04-14
add a lib for juju topology

simskij:
## Issue
This PR adds a common library for JujuTopology. It strips away all of the complexity we've added in other charms in favor of a simple and coherent API.  

## Solution
- Get rid of the abstract base class.
- Get rid of the mix between 
- Add the ability to exclude items from the dict
- Add the ability to remap keys in the dict

## Context
N/A

## Testing Instructions
- Try the library with one of our existing charms, replacing the current implementation with the one in the library.
- Run all tests, both unit and integration. 
- Perform manual sanity checks

## Release Notes
Adds a reusable library for creating, generating and formatting Juju topology.


lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,230 @@
...
+        Keeping the dictionary ordered is important to be able to
+        compare dicts without having to resort to deep comparisons.
+
+        Args:
+            remapped_keys: A dictionary mapping old key names to new key names,
+                which will be substituted when invoked.
+            excluded_keys: A list of key names to exclude from the returned dict.
+            uuid_length: The length to crop the UUID to.
+        """
+        model_uuid = self.model_uuid[: uuid_length - 1] if uuid_length else self.model_uuid
```
sed-i, line 174: - Need to account for:
  - uuid_length = int(0)
  - negative values
- The `-1` is not needed.

```suggestion
        uuid_length = uuid_length or len(self.model_uuid)
        model_uuid = self.model_uuid[: uuid_length]
```

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,230 @@
...
+                ret.pop(exclusion)
+
+        if remapped_keys:
+            ret = OrderedDict(
+                (remapped_keys.get(k), v) if remapped_keys.get(k) else (k, v) for k, v in ret.items()  # type: ignore
+            )
+
+        return ret
+
+    def as_identifier(self) -> str:
```
sed-i, line 196: `excluded_keys=["unit", "charm_name"]` and `uuid_length=8` is very opinionated.
Is there a better name for this than `identifier`, e.g. `app identifier`? Or perhaps just let users extend? Or perhaps:

```suggestion
    def app_identifier(self, excluded_keys=("unit", "charm_name"), uuid_length=8) -> str:
```

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,230 @@
...
+            application=data["application"],
+            unit=data.get("unit", ""),
+            charm_name=data.get("charm_name", ""),
+        )
+
+    def as_dict(
+        self,
+        *,
+        remapped_keys: Dict[str, str] = {},
+        excluded_keys: List[str] = [],
```
sed-i, line 160: Currently there is no issue, but technically this should be
```suggestion
        remapped_keys: Dict[str, str] = None,
        excluded_keys: List[str] = None,
```
cf. mutable default argument.

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,230 @@
...
+        anything but other identifiers.
+        """
+        values = self.as_dict(
+            excluded_keys=["unit", "charm_name"],
+            uuid_length=8,
+        ).values()
+
+        return "_".join([str(val) for val in values]).replace("/", "_")
+
+    def as_relabelled_dict(self):
```
sed-i, line 209: I would drop this function in favor of using `as_dict` with a new additional arg `key_mutator=lambda k: "juju_" + k`.

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,230 @@
...
+        the leader unit (ie. the unit that produced the dict).
+        """
+        items = self.as_dict(
+            remapped_keys={"charm_name": "charm"},
+            excluded_keys=["unit"],
+        ).items()
+
+        return {"juju_{}".format(key): value for key, value in items if value}
+
+    def as_label_matchers(self) -> str:
```
sed-i, line 222: I would drop this function in favor of using `app_identifier` with a new additional arg `separator=', '`

tests/test_juju_topology.py
```diff
@@ -0,0 +1,83 @@
...
+        self.topology = JujuTopology(
+            self.input["model"],
+            self.input["model_uuid"],
+            self.input["application"],
+            self.input["unit"],
+            self.input["charm_name"],
+        )
+
+    def test_dict_includes_all_expected_keys(self):
+        assert self.topology.as_dict() == self.input
```
rbarry82, line 35: Pretty much all of these should be replaced with `unittest` assertions.

For example:
```python
self.assertEqual(self.topology.as_dict(), self.input)
```

It gives nicer results if the test fails, follows common practices, and gets you a bunch of stuff "free", like `assertContains`

Plus, using these is a good assurance that the cleanup steps (for the harness, for example, if it's ever used here) actually run.

tests/test_juju_topology.py
```diff
@@ -0,0 +1,83 @@
...
+            ("juju_{}".format("charm" if key == "charm_name" else key), val)
+            for key, val in self.input.items()
+        )
+        expected.pop("juju_unit")
+        assert self.topology.as_relabelled_dict() == expected
+
+    def test_label_matchers_creates_a_valid_matcher(self):
+        prefixed = OrderedDict(
+            ("juju_{}".format("charm" if raw_key == "charm_name" else raw_key), raw_val)
+            for raw_key, raw_val in self.input.items()
```
rbarry82, line 66: `for raw_key, raw_val in self.input.items() if raw_key not in ["juju_unit"]`

As an aside, using one character identifiers in comprehensions is completely Pythonic, and `... for k, v in dict.items()` is idiomatic.

This could quickly get out of hand, so maybe:

```python
def _filter_dict(from: dict, filtered_keys: List):
    return {k: v for k, v in dict.items() if k not in filtered_keys}
```

Then you could use this instead of `.pop()`, and it won't be a "test case" since it doesn't start with `test_`

lib/charms/observability_libs/v0/juju_topology.py
simskij: If the naming is bad, I'm all for renaming it, but I do agree with Ryan that we should be opinionated. Any flexibility should be out of necessity, not because it is nice to have.

lib/charms/observability_libs/v0/juju_topology.py
simskij: Yeah, I'm totally fine with making them properties instead. I wanted to make the API consistent, and opted for dropping properties altogether. I can just as well see us going the other direction and making all methods except the class methods `@properties`, but then I think it would make sense to also drop the `as_` prefix.

lib/charms/observability_libs/v0/juju_topology.py
simskij: That is very true. Let's verify the uuid in the constructor and drop the defensive code here.

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,230 @@
...
+
+        Args:
+            data: a dictionary with five keys providing topology information. The keys are
+                - "model"
+                - "model_uuid"
+                - "application"
+                - "unit"
+                - "charm_name"
+                `unit` and `charm_name` may be empty, but will result in more limited
+                labels. However, this allows us to support payload-only charms.
```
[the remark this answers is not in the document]
simskij: No, it should probably be "charms without workloads". Could be an integrator charm, but could also be a charm fully implemented in the charm layer.

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,230 @@
...
+from typing import Dict, List, Optional
+
+# The unique Charmhub library identifier, never change it
+LIBID = "bced1658f20f49d28b88f61f83c2d232"
+
+LIBAPI = 0
+LIBPATCH = 1
+
+
+class JujuTopology:
```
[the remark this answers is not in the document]
simskij: > I imagine that for a juju dev "juju topology" would mean something very different.

If it does, we have an issue. The term was, to my knowledge, coined in observability by @mmanciop, but it should be valid regardless of subcontext as long as we still talk about Juju deployments.

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,230 @@
...
+```python
+scrape_metadata = json.loads(relation.data[relation.app].get("scrape_metadata", "{}"))
+topology = JujuTopology.from_dict(scrape_metadata)
+```
+
+### Using the class constructor
+
+Enables instantiation using whatever values you want. While this
+is useful in some very specific cases, this should not be considered
+the least favored approach.
```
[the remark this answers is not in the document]
simskij: the `not` that snuck it's way into that sentence definitely has to go. I like @rbarry82's suggestion about adding something about why you likely don't want to use it.

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,230 @@
...
+class JujuTopology:
+    """JujuTopology is used for storing, generating and formatting juju topology information."""
+
+    def __init__(
+        self,
+        model: str,
+        model_uuid: str,
+        application: str,
+        unit: Optional[str] = "",
+        charm_name: Optional[str] = "",
```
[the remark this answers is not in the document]
simskij: Great feedback! I'll try them out and see which one I prefer. 😅

lib/charms/observability_libs/v0/juju_topology.py
simskij: That's an excellent catch. Will fix.

lib/charms/observability_libs/v0/juju_topology.py
simskij: I prefer readability. It's not like the methods are doing much, so we might as well keep it. Not a super fan of adding additional arguments unless it is unavoidable.

tests/test_juju_topology.py
simskij: I like how you think! Will switch it out.

lib/charms/observability_libs/v0/juju_topology.py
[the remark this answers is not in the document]
sed-i: > The goal is less to classify, and more to identify

Totally agree, that's why I don't like `Taxonomy` for this.
But JujuTopolgy does not translate in my mind into "identifying app/unit". Nor does `CharmTopology`.
I.e. "network topology" is not analogous to "juju topology" the way we currently use it.

lib/charms/observability_libs/v0/juju_topology.py
sed-i: The defensive part here is for the incoming `uuid_length` arg, not the constructor arg.

lib/charms/observability_libs/v0/juju_topology.py
sed-i: This lib is called observability_libs so we can totally do our opinionated contracts here.
However, it may hamper the adoptability of this lib outside of o11y charms.

lib/charms/observability_libs/v0/juju_topology.py
sed-i: In that case... `ObservabilityTopology`? Or just drop "Juju" and call this "Topology" (as it comes from observability_libs).

lib/charms/observability_libs/v0/juju_topology.py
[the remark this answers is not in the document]
rbarry82: It is definitely possible that it's one of those things which comes about when you have to manage an environment which is unnecessarily large, and naming things after planets/mythology/whatever doesn't work anymore. 

"I got an alert. How do I can I tell from the information I already have where it's located, what it does, and a guess at how many of these there are so I know what the impact is going to be?"

It's kinda "topology as seen by Juju", since we're not adding the controller name (which isn't exposed in any meaningful way in OF or in Juju anyway -- see [here](https://github.com/juju/juju/blob/ebaa903b73f02ff51a7386274b8de1cd292abd79/worker/uniter/runner/context/context.go#L1067) for a more or less list of what could conceivably be added), but it's still in the sense of "this is how things are arranged."

The name is sort of aiming for understanding from ops/admins rather than devs.

lib/charms/observability_libs/v0/juju_topology.py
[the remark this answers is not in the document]
rbarry82: A charm fully implemented in the charm layer would still have a unit, but point taken.

lib/charms/observability_libs/v0/juju_topology.py
rbarry82: What I meant was that the UUID is cut down to size both here and in `as_identifier`, the only two places where it's referenced.

We could try to check for zero length others in both methods, or we could say  `model_uuid` is a required argument which must actually be a UUID. If it doesn't walk like a UUID and talk like a UUID when you instantiate the object, you don't get a Topology.

This ensures that UUIDs are actually UUIDs and not anything like `zyxwtsrq-woo!` so it means what the attribute says it means, and it also means that we can move these checks (length, negative values, etc) to the constructor where we ensure this. We can also trim it down there so it doesn't need to be handled elsewhere, and it's immediately clear to anyone reading the code exactly why their "UUID" comes out as 8 characters in the label.

Since Python can't reasonably enforce types at runtime (especially with 3.5 compatibility), "type checking" in the constructor with these guards (or more reasonably, trying to pass it to `uuid.UUID` and seeing what we get back) is as good as it gets. If that check passes, we don't need to check here, because we know for sure what it is. If that check fails, they don't get a topology object.

To be defensive at that point, making it a read-only property so it can't be changed would work, but it's Python, so they could always just mutate it anyway even with `__`), so we rely on being good citizens.

lib/charms/observability_libs/v0/juju_topology.py
rbarry82: Sure, it's observability libs, but we also want to try to get this into "core" OF, and a good contract will help there.

Beyond that, we want to make sure that the things we are actually observing set their labels in a way that they're observable, right? A very strict contract is ideal there, and this contract is not strictly tied to observability except in the sense that users may want to observe their charms, and these are useful labels. It could be equally useful for geo-replicating Ceph/Gluster, determining locality for distributed databases, routing endpoints to closest-hop like BGP, etc. 

Imagine an "integrator charm" into which an admin can pass a dict/map which correlates controllers/clouds/AZs/datacenters/whatever and models so a highly distributed database/object store/whatever can effectively route clients to the closest endpoint. This class can provide that, and it is not specific to observability. Will anyone actually do this? No idea. But keeping a very strict definition of what these values mean will certainly help, since users won't have have to try to guess what values are in "topology" from a charm they didn't write.

I suggested `@property` mostly because then you don't need `()` to call a method which takes no arguments of any kind, but if `model_uuid`, for example, was also:

```python
@property
def model_uuid(self) -> str:
    return self._model_uuid
```

It would also make it more challenging to accidentally (or intentionally) reassign it. Sure, you could just  `topology._model_uuid = ...`, but "Pythonic" practice is basically "any `attr` (which includes methods) which is not prepended by `_` is public". 

I mean, it's actually public anyway, but think of it like exporting a field from a struct in Go. If you intend for it to be "internal", add `_`, or it will definitely pop up in autocompletion for anything which uses LSP (pycharm, emacs, jedi, neovim, vscode, etc), which makes the surface area we want to expose muddier.

lib/charms/observability_libs/v0/juju_topology.py
rbarry82: Mentioned on Meet, but we want this to go into OF if we can. Even if it ultimately gets exposed as `charm.topology = FooBarBazTopology.from_charm(self)`, we may as well wait to agonize over an appropriate name until those conversations start to converge.

lib/charms/observability_libs/v0/juju_topology.py
rbarry82: It is an excellent catch.

lib/charms/observability_libs/v0/juju_topology.py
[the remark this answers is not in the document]
balbirthomas: Taxonomy implies "classification" of charms, Topology implies geometric arrangement. I think "Juju Topology" is more appropriate since the topology information is used to provide information of the location (geometric property) i.e. source of the metrics, etc.

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,230 @@
...
+        """
+        return cls(
+            model=data["model"],
+            model_uuid=data["model_uuid"],
+            application=data["application"],
+            unit=data.get("unit", ""),
+            charm_name=data.get("charm_name", ""),
+        )
+
+    def as_dict(
```
balbirthomas, line 156: Since "Juju Topology" conceptually is a "mapping" would it be more appropriate to [emulate the mapping type](https://docs.python.org/3/reference/datamodel.html?emulating-container-types#emulating-container-types) rather than provide a function to return a mapping ?

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,230 @@
...
+            ]
+        )
+
+        for exclusion in excluded_keys:
+            if exclusion in ret:
+                ret.pop(exclusion)
+
+        if remapped_keys:
+            ret = OrderedDict(
+                (remapped_keys.get(k), v) if remapped_keys.get(k) else (k, v) for k, v in ret.items()  # type: ignore
```
balbirthomas, line 191: If I remember correctly the need for  "remapped keys" is an accident of the way in which the code consume Juju Topology was implemented. That being the case it may be prudent to not carry forward this feature.

lib/charms/observability_libs/v0/juju_topology.py
balbirthomas, line 196: This method is very specific to a few observability charms. If the intention is to create a generic topology object that may some day reside in the Operator Framework then perhaps we should not include such a method in the topology object. It is a one liner to convert a topology into an identifier (particularly if the former has the behaviour of an ordered dict).

lib/charms/observability_libs/v0/juju_topology.py
balbirthomas: I am also in favour of dropping this method and refactoring related code to always having `juju_` prefix.

lib/charms/observability_libs/v0/juju_topology.py
balbirthomas: I am also in favour of dropping this code and instead ensuring that `from_charm()` either
- always includes unit (the unit that constructed it)
- never includes unit (user of class has to add it if they need to)
Juju Topology is doing too many things and trying to satisfy too many use cases. This makes its interface awkward and difficult to use.

lib/charms/observability_libs/v0/juju_topology.py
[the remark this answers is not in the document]
simskij: I'm killing this conversation. We're not renaming it. 😅

lib/charms/observability_libs/v0/juju_topology.py
simskij: That is not an option, unfortunately. Adding a `key_mutator` function further limits the usability of having this library in the first place (ie. not having to think too much about how it *should* be)

lib/charms/observability_libs/v0/juju_topology.py
simskij: I disagree. It currently has one method and 3 properly documented properties. The reason why we don't want to always have it prefixed with `juju_` is simple: that is not how the charm itself sees it, so then we'd have no way of reconstructing the original data fed to the factory or constructor.

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,230 @@
...
+
+Enables instantiation by supplying the charm as an argument. When
+creating topology objects for the current charm, this is the recommended
+approach.
+
+```python
+topology = JujuTopology.from_charm(self)
+```
+
+### Using the `from_dict` method
```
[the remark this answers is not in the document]
simskij: The reason why I opted for not naming it `from_relation` is because that already isn't the only way we use it. Consider for instance labels in alert rules.

lib/charms/observability_libs/v0/juju_topology.py
simskij: See comment above about being able to reconstruct the data.

lib/charms/observability_libs/v0/juju_topology.py
simskij: The one-liner argument falls on the fact that we are already struggling to keep it in sync across charms.

lib/charms/observability_libs/v0/juju_topology.py
simskij: Seems like that would add unnecessary complexity, as opposed to just keeping it a simple intermediary object.

lib/charms/observability_libs/v0/juju_topology.py
simskij: You can have a look at how I addressed it. I think it takes care of most concerns.

lib/charms/observability_libs/v0/juju_topology.py
simskij: Implemented the getters.

jnsgruk:
This could also do with a slight adjustment to the top-level README to add a reference to this lib :)

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,281 @@
...
+        return ret
+
+    @property
+    def identifier(self) -> str:
+        """Format the topology information into a terse string.
+
+        This crops the model UUID, making it unsuitable for comparisons against
+        anything but other identifiers. Mainly to be used as a display name or file
+        name where long strings might become an issue.
+        """
```
sed-i, line 218: Seems like a suitable candidate for a [doctest](https://docs.python.org/3/library/doctest.html).

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,281 @@
...
+            excluded_keys=["unit", "charm_name"],
+        )
+
+        parts["model_uuid"] = self.model_uuid_short
+        values = parts.values()
+
+        return "_".join([str(val) for val in values]).replace("/", "_")
+
+    @property
+    def prefixed_dict(self):
```
sed-i, line 229: We should come up with a proper term for this - in my mind, `prefixed` does not imply "no unit label".
That being said, this is under o11y libs so it's not a big deal we have an implicit use case in mind here.

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,281 @@
...
+        """Format the topology information into a dict with keys having 'juju_' as prefix.
+
+        Relabelled topology never includes the unit as it would then only match
+        the leader unit (ie. the unit that produced the dict).
+        """
+        items = self.as_dict(
+            remapped_keys={"charm_name": "charm"},
```
sed-i, line 236: We should start a list for the time we choose to break backwards compat and add there to either "charm" or "charm_name" everywhere.

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,281 @@
...
+            excluded_keys=["unit"],
+        ).items()
+
+        return {"juju_{}".format(key): value for key, value in items if value}
+
+    @property
+    def label_matchers(self) -> str:
+        """Format the topology information into a verbose string.
```
sed-i, line 244:
```suggestion
        """Format the topology information into a promql/logql label matcher string.
```

tests/test_juju_topology.py
```diff
@@ -0,0 +1,96 @@
...
+
+        self.assertEqual(self.topology.prefixed_dict, expected)
+
+    def test_label_matchers_creates_a_valid_matcher(self):
+        prefixed = OrderedDict(
+            ("juju_{}".format("charm" if raw_key == "charm_name" else raw_key), raw_val)
+            for raw_key, raw_val in self.input.items()
+        )
+        prefixed = _filter_dict(prefixed, ["juju_unit"])
+        expected = ", ".join('{}="{}"'.format(key, value) for key, value in prefixed.items())
```
sed-i, line 78: This is a re-implementation of `label_matchers`.
Wouldn't it make more sense to have a hard-coded literal here instead?

sed-i (approved):

tests/test_juju_topology.py
simskij: Good catch, switched it out

lib/charms/observability_libs/v0/juju_topology.py
simskij: Done

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,289 @@
...
+    def is_valid_uuid(self, uuid):
+        """Validates the supplied UUID against the Juju Model UUID pattern."""
+        regex = re.compile(
+            "^[a-f0-9]{8}-?[a-f0-9]{4}-?4[a-f0-9]{3}-?[89ab][a-f0-9]{3}-?[a-f0-9]{12}$"
+        )
+        return bool(regex.match(uuid))
+
+    @classmethod
+    def from_charm(cls, charm):
+        """Creates a JujuTopology instance by using the model data available on a charm object.
```
dstathis, line 138: I would consider making this the default constructor. I believe it is used most and it is certainly the most convenient.

lib/charms/observability_libs/v0/juju_topology.py
dstathis: I agree here. Something like `alert_labels_dict` would be much better.

lib/charms/observability_libs/v0/juju_topology.py
simskij: It seems to me like it's less confusing if the constructor is as straight-forward as possible (i.e. by asking the user to provide the topology components one by one as arguments). I get your point and agree with the convenience argument however.

lib/charms/observability_libs/v0/juju_topology.py
simskij: Renamed to `label_matcher_dict`.

README.md
```diff
@@ -15,6 +15,8 @@ Where this charm should be used, is to access one of the following libraries dur
 - [KubernetesServicePatch] - a small library used to patch the Juju auto-created Kubernetes Service
   during the deployment of a sidecar charm to contain the correct ports for an application.
 
+- [JujuTopology] - Used to create and output Juju topologies either from charms, relation data, or parts.
```
jnsgruk, line 18: This reference style link needs to be defined at the bottom of the file, see line 25-27.

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,289 @@
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+"""## Overview.
+
+This document explains how to use the JujuTopology class to
```
jnsgruk, line 5:
```suggestion
This document explains how to use the `JujuTopology` class to
```

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,289 @@
...
+create and consume topology information in a consistent manner.
```
jnsgruk, line 6:
```suggestion
create and consume topology information from Juju in a consistent manner.
```

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,289 @@
...
+
+The goal of the Juju topology is to uniquely identify a piece
+of software running across any of your Juju-managed deployments.
+This is achieved by combining the following four elements:
+
+- Model name
+- Model UUID
+- Application Name
+- Unit Identifier
```
jnsgruk, line 15:
```suggestion
- Application name
- Unit identifier
```

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,289 @@
...
+
+and in some cases, the charm name.
```
jnsgruk, line 17: "in some cases" probably needs some clarification

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,289 @@
...
+
+For a more in-depth description of the concept, as well as a
+walk-through of it's use-case in observability, see
+[this blog post](https://juju.is/blog/model-driven-observability-part-2-juju-topology-metrics)
+on the Juju blog.
+
+## Library Usage
+
+This library may be used to create and consume JujuTopology objects.
```
jnsgruk, line 26:
```suggestion
This library may be used to create and consume `JujuTopology` objects.
```

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,289 @@
...
+The JujuTopology class provides three ways to create instances:
```
jnsgruk, line 27:
```suggestion
The `JujuTopology` class provides three ways to create instances:
```

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,289 @@
...
+        model: str,
+        model_uuid: str,
+        application: str,
+        unit: str = None,
+        charm_name: str = None,
+    ):
+        """Build a JujuTopology object.
+
+        A `JujuTopology` object is used for storing and transforming
+        Juju Topology information. This information is used to
```
jnsgruk, line 105:
```suggestion
        Juju topology information. This information is used to
```

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,289 @@
...
+            excluded_keys=["unit", "charm_name"],
+        )
+
+        parts["model_uuid"] = self.model_uuid_short
+        values = parts.values()
+
+        return "_".join([str(val) for val in values]).replace("/", "_")
+
+    @property
+    def label_matcher_dict(self):
```
jnsgruk, line 237: This, and a few of the other properties could benefit from a type annotation on the return type.

jnsgruk (dismissed):
This looks good to me now. Comments are all nits on documentation and nothing concrete, can be punted into another PR if you'd prefer.

README.md
simskij: Addressed. Thanks!

lib/charms/observability_libs/v0/juju_topology.py
simskij: It can go altogether for now, as these are currently hypothetical scenarios.

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,288 @@
...
+
+# The unique Charmhub library identifier, never change it
+LIBID = "bced1658f20f49d28b88f61f83c2d232"
+
+LIBAPI = 0
+LIBPATCH = 1
+
+
+class InvalidUUIDError(Exception):
+    """Creates a raisable exception whenever an invalid UUID is provided."""
```
sed-i, line 83: Most common exception docstring style I encountered is:

```suggestion
    """Invalid UUID was provided."""
```
or
```suggestion
    """Raised when an invalid UUID is provided."""
```

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -0,0 +1,288 @@
...
+        self._application = application
+        self._charm_name = charm_name
+        self._unit = unit
+
+    def is_valid_uuid(self, uuid):
+        """Validates the supplied UUID against the Juju Model UUID pattern."""
+        regex = re.compile(
+            "^[a-f0-9]{8}-?[a-f0-9]{4}-?4[a-f0-9]{3}-?[89ab][a-f0-9]{3}-?[a-f0-9]{12}$"
+        )
+        return bool(regex.match(uuid))
```
sed-i, line 133: Should we use the builtin [`uuid`](https://docs.python.org/3/library/uuid.html) here? The builtin func also ignores braces and hyphens, so wider compatibility.

```suggestion
        # from uuid import UUID
        try:
            UUID(uuid)
        except ValueError:
            return False
        return True
```

sed-i (approved):

lib/charms/observability_libs/v0/juju_topology.py
simskij: We can do that at a later stage if we feel it provides enough value. Fwiw regex this also supports both with and without hyphens (see `-?`).

PR #13, opened 2022-06-10
Replace leader guard with is_patched guard

sed-i:
## Issue
The k8s patching had a leader guard inside on-install, which always happens before leader-elected.
It's possible that juju already elected a leader by the time `install` fires, but it is technically incorrect to assume so.
A unit does not need to be a leader to patch k8s.

## Solution
Use `is_patched` guard instead of leader guard.

## Context
This is only a partial solution so far, because we still get these kind of errors:
```
ERROR juju.worker.caasapplicationprovisioner.runner exited "prometheus-k8s": Operation cannot be fulfilled on pods "prometheus-k8s-0": the object has been modified; please apply your changes to the latest version and try again
```

## Testing Instructions
Deploy e.g. prometheus with this version of the lib.

## Release Notes
Replace leader guard with is_patched guard in k8s patching.

### DIFF
--- a/lib/charms/observability_libs/v1/kubernetes_service_patch.py
+++ b/lib/charms/observability_libs/v1/kubernetes_service_patch.py
@@ -125,7 +125,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
+LIBPATCH = 2
 
 ServiceType = Literal["ClusterIP", "LoadBalancer"]
 
@@ -232,7 +232,7 @@ class KubernetesServicePatch(Object):
         Raises:
             PatchFailed: if patching fails due to lack of permissions, or otherwise.
         """
-        if not self.charm.unit.is_leader():
+        if self.is_patched():
             return
 
         client = Client()
--- a/tests/test_kubernetes_service.py
+++ b/tests/test_kubernetes_service.py
@@ -373,13 +373,17 @@ class TestK8sServicePatch(unittest.TestCase):
 
     @patch(f"{MOD_PATH}.Client.patch")
     @patch(f"{MOD_PATH}.ApiError", _FakeApiError)
+    @patch(f"{CL_PATH}.is_patched", lambda *args: False)
     @patch("lightkube.core.client.GenericSyncClient", Mock)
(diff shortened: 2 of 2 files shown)


Abuelodelanada (approved):

jnsgruk (approved):

jnsgruk:
Merged, thanks. I think you should have rights to publish this to Charmhub @sed-i?

sed-i:
So needing to add a `patch` to the unittests _here_ was an early indication that this change is going to break people's CI:
```
lightkube.core.exceptions.ConfigError: [Errno 2] No such file or directory: '/var/run/secrets/kubernetes.io/serviceaccount/token'
lightkube.core.exceptions.ConfigError: Configuration file ~/.kube/config not found
```
Revert and create v2?

jnsgruk:
Oh interesting. Yeh, I'm away from my computer for a bit, but if it's breaking things then definitely revert - where did you catch it?

I'm interested because in most places I've seen this, the whole class is mocked out for tests?

sed-i:
> where did you catch it

Prometheus unit tests.

sed-i:
Prometheus unit tests [mock methods](https://github.com/canonical/prometheus-k8s-operator/blob/e5b50c94bab6d2a5ff549e571bb64c4ffd3b008e/tests/unit/test_remote_write.py#L259), not the entire class, so reverting is best.
Created #14 for this.

rbarry82:
> So needing to add a `patch` to the unittests _here_ was an early indication that this change is going to break people's CI:
> 
> ```
> lightkube.core.exceptions.ConfigError: [Errno 2] No such file or directory: '/var/run/secrets/kubernetes.io/serviceaccount/token'
> lightkube.core.exceptions.ConfigError: Configuration file ~/.kube/config not found
> ```
> 
> Revert and create v2? (Haven't published to charmhub yet.)

Does this actually break "real" code where k8s/microk8s is deployed, or just unit tests?

Since `lightkube` itself wasn't updated, the "right" way may be catching the exception. It's happening now because `Client` isn't actually instantiated until `patch|is_patched`, but:

```python
from lightkube.core import exceptions

...

class KubernetesServicePatch(Object):
    def _patch(self, ...):
        try:
            _client = Client()
        except exceptions.ConfigError:
            logger.warning(...)
            return
        if self.is_patched(client):
            ...
```

PR #14, opened 2022-06-10
Absorb exception in _patch

sed-i:
## Issue
#13 introduced a change that would break people's CI. 

## Solution
Per [suggestion from @rbarry82 in #13](https://github.com/canonical/observability-libs/pull/13#issuecomment-1152758809), absorb the exception in _patch that is raised when a client fails to instantiate.

## Context
NTA.

## Testing Instructions
- Deploy e.g. prometheus with this version of the lib.
- Run CI.

## Release Notes
Absorb exception in _patch.


lib/charms/observability_libs/v1/kubernetes_service_patch.py
```diff
@@ -59,7 +59,7 @@ def __init__(self, *args):
 
 ```python
 # ...
-from charms.observability_libs.v0.kubernetes_service_patch import KubernetesServicePatch
+from charms.observability_libs.v1.kubernetes_service_patch import KubernetesServicePatch
```
rbarry82, line 62: I suppose the question is still "is this enough of an API change to really need v1"? It seems like it's still the same for consumers and is just a bugfix

sed-i: It's already v1. This is just a doc fix.

Abuelodelanada (approved):

PR #19, opened 2022-06-21
Add k8s resource limit patch lib

sed-i:
## Issue
Need to be able to limit resource usage of a charm.

Crossref: [OPENG-272](https://warthogs.atlassian.net/browse/OPENG-272)

## Solution
- Create lib to patch k8s compute resource `limits` and `requests`.
- Turn the placeholder charm into a tester charm, and use it in itests.

## Context
NTA.

## Testing Instructions
- Deploy [prom/318](https://github.com/canonical/prometheus-k8s-operator/pull/318).
- Compare to `kubectl patch statefulset prom -n welcome -p '{"spec": {"template": {"spec": {"containers": [{"name":"prometheus", "resources": {"limits": {"cpu": "2", "memory":"2Gi"}, "requests": {"cpu": "2", "memory": "1Gi"}} }] }}}}'`

## Release Notes
Add k8s resource limit patch lib.


lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -0,0 +1,302 @@
...
+            charm: the charm that is instantiating the library.
+            container_name: the container for which to apply the resource limits.
+            limits: a dictionary for `limits` resources.
+            requests: a dictionary for `requests` resources.
+            refresh_event: an optional bound event or list of bound events which
+                will be observed to re-apply the patch.
+        """
+        super().__init__(charm, "kubernetes-compute-resource-patch")
+        self.charm = charm
+        self.container_name = container_name
```
Abuelodelanada, line 140: Shouldn't these properties be private?

Abuelodelanada (approved):

sed-i:
There's a complication:
- When patching the StatefulSet with `{"memory": "0.9Gi"}`, k8s generates a PodSpec with `{"memory": "966367641600m"}`, i.e. millibytes (e.g. to convert it back to GiB, `value = bitmath.Byte(float(memory[:-1]) / 1000).to_GiB().value`).
- K8s docs [claim](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/) that the `m` suffix if part of [SI](https://physics.nist.gov/cuu/Units/binary.html), but that seems to be false, at least for binary multiples.

Is there a way to force K8s to KEEP the same unit the user provided? @simskij @rbarry82

simskij:
> There's a complication:
> 
> * When patching the StatefulSet with `{"memory": "0.9Gi"}`, k8s generates a PodSpec with `{"memory": "966367641600m"}`, i.e. millibytes (e.g. to convert it back to GiB, `value = bitmath.Byte(float(memory[:-1]) / 1000).to_GiB().value`).
> * K8s docs [claim](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/) that the `m` suffix if part of [SI](https://physics.nist.gov/cuu/Units/binary.html), but that seems to be false, at least for binary multiples.
> 
> The `.is_ready()` method compares statefulset to actual pod and currently it is buggy because of the above.
> 
> Is there a way to force K8s to KEEP the same unit the user provided? @simskij @rbarry82

Not that I know off, but I'll dig. What happens if you instead of 0.9Gi set it to 966367641600m from the get-go? Does it still bug?

simskij:
> > There's a complication:
> > 
> > * When patching the StatefulSet with `{"memory": "0.9Gi"}`, k8s generates a PodSpec with `{"memory": "966367641600m"}`, i.e. millibytes (e.g. to convert it back to GiB, `value = bitmath.Byte(float(memory[:-1]) / 1000).to_GiB().value`).
> > * K8s docs [claim](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/) that the `m` suffix if part of [SI](https://physics.nist.gov/cuu/Units/binary.html), but that seems to be false, at least for binary multiples.
> > 
> > The `.is_ready()` method compares statefulset to actual pod and currently it is buggy because of the above.
> > Is there a way to force K8s to KEEP the same unit the user provided? @simskij @rbarry82
> 
> Not that I know off, but I'll dig. What happens if you instead of 0.9Gi set it to 966367641600m from the get-go? Does it still bug?

Ok, so found it. The reason for this is that 0.9GiB (Gibibyte) in its canonical form (ie. without fractional digits, using the largest possible suffix, is `m`.

> Before serializing, Quantity will be put in "canonical form". This means that Exponent/suffix will be adjusted up or down (with a corresponding increase or decrease in Mantissa) such that: a. No precision is lost b. No fractional digits will be emitted c. The exponent (or suffix) is as large as possible. The sign will be omitted unless the number is negative.

If we just disallow setting fractions in the first place, the problem will be solved. For instance, 900Mi will not get converted, while 0.9Gi will. Likely, 900Mi was also what the user tried to express when they put in 0.9Gi, which isn't entirely true as 0.9Gi = 6866,46Mi.

sed-i:
> disallow setting fractions in the first place, the problem will be solved.

There's also the case of `1000000Mi` we should probably support what k8s supports and do proper conversion.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
sed-i: Done in next commit.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -0,0 +1,409 @@
...
+        return patched and ready
+
+    @classmethod
+    def _get_container(cls, container_name: str, containers: List[Container]) -> Container:
+        """Find our container from the container list, assuming list is unique by name.
+
+        Typically, *.spec.containers[0] is the charm container, and [1] is the (only) workload.
+
+        Raises:
+            StopIteration, if the user-provided container name does not exist in the list.
```
dstathis, line 357: Perhaps a more appropriate exception should be raised?

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -0,0 +1,409 @@
...
+        return self._charm.app.name
+
+    @property
+    def _pod(self) -> str:
+        """Name of the unit's pod.
+
+        Returns:
+            str: A string containing the name of the current unit's pod.
+        """
+        return "-".join(self._charm.unit.name.rsplit("/", 1))
```
dstathis, line 394:
```suggestion
        return self._charm.unit.name.replace("/", "-", 1)
```

requirements.txt
```diff
@@ -3,4 +3,5 @@
 
 git+https://github.com/canonical/operator/#egg=ops
 PyYAML
-lightkube
\ No newline at end of file
+lightkube
+bitmath == 1.3.3.1
```
dstathis, line 7: Why are we pinning this?

dstathis (approved):

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -0,0 +1,409 @@
...
+            )
+
+    @classmethod
+    def _conv_res_req(cls, res_req: ResourceRequirements) -> ResourceRequirements:
+        """Convert ResourceRequirements to comparable form.
+
+        - Convert "memory" to GiB.
+        - Convert "cpu" to float.
+
+        When patching the StatefulSet with {"memory": "0.9Gi"}, the actual PodSpec has
```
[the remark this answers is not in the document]
sed-i: Both Gi and GiB would work for the charm. Here in the comment I'm sticking to K8s representation (they omit the B) for consistency.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
sed-i: The suggestion is opposite to the original:
```python
>>> "one/two/0".replace("/", "-", 1)
'one-two/0'
>>> "-".join("one/two/0".rsplit("/", 1))
'one/two-0'
```

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
dstathis: ValueError or a custom ContainerNotFoundError

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
dstathis: Ah I see the difference. Good catch.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
simskij: > ContainerNotFoundError

I prefer this one

Abuelodelanada (dismissed):

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
dstathis: I feel that ContainerNotFound *would* likely be more helpful when encountered in the logs. It would immediately tell you what the issue is (or at least help narrow it down).

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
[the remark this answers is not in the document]
dstathis: It just looked a bit weird given two lines above but not at all a deal breaker if you prefer this.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -0,0 +1,412 @@
...
+        {"memory": "966367641600m"}; similarly, {"cpu": "0.30000000000000004"} -> {"cpu": "301m"}.
+        So need to parse the strings and convert before comparing.
+        """
+
+        def _conv(dct: ResourceSpecDict) -> Optional[ResourceSpecDict]:
+            """Convert the memory value of a ResourceSpecDict to GiB representation.
+
+            Raises:
```
balbirthomas, line 283: The docstring says "raises ValueError for invalid input". However immediate after this method returns `None` if `dct` is falsy. This was surprising. Why not just raise `ValueError` for any invalid input ? This may also allow you to always return a single return type.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -0,0 +1,412 @@
...
+            if cpu := copy.get("cpu"):
+                # TODO need to take into account m, k/K, M, G, T, P and E, but for CPU count it's
+                # probably ok to only support "m" and plain decimals for now.
+                # https://github.com/gtsystem/lightkube/issues/36
+                if cpu.endswith("m"):
+                    # This is milli. Divide by 1000.
+                    value = int(cpu[:-1]) / 1000.0
+                else:
+                    # Round up to whole millis (e.g. 0.30000000000000004 -> 0.301)
+                    value = ceil(float(cpu) * 1000) / 1000.0
```
balbirthomas, line 316: Any reason not to use Python's own `round()` function ?

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -0,0 +1,412 @@
...
+                    )
+                copy["memory"] = f"{str(value)}Gi"
+
+            if cpu := copy.get("cpu"):
+                # TODO need to take into account m, k/K, M, G, T, P and E, but for CPU count it's
+                # probably ok to only support "m" and plain decimals for now.
+                # https://github.com/gtsystem/lightkube/issues/36
+                if cpu.endswith("m"):
+                    # This is milli. Divide by 1000.
+                    value = int(cpu[:-1]) / 1000.0
```
balbirthomas, line 313: May want to extend that comment above to explicitly state if loss of precision is intended, as it seems.

balbirthomas (dismissed):

dstathis (approved):

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
sed-i: `None` is actually valid in this context: it means no limits.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -0,0 +1,412 @@
...
+                ValueError, for invalid input.
+            """
+            if not dct:
+                return None
+            copy = dct.copy()
+
+            if memory := copy.get("memory"):
+                if memory.endswith("m"):
+                    # This is milli. Divide by 1000.
+                    value = bitmath.Byte(float(memory[:-1]) / 1000).to_GiB().value
```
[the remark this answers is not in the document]
sed-i: (Note: it's `bitmath.Byte`, whose `.value` is still a float)
Yes, there could be some unexpected differences because what really needs to be done is round up to 0.001.
I do not plan to merge this before replacing it with lightkube's `parse_quantity` (review pending).

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
sed-i: Yes, K8s spec is round up to millis. But all this will go away when using `parse_quantity`.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
sed-i: Changed to custom exception.

sed-i:
Problem: when there is a single unit of prometheus and the user sets too high resource limits, then juju is stuck in 
```
prometheus-k8s/0  unknown   lost
```
because the pod cannot be scheduled and there is no charm to take `juju config` changes to repatch the statefulset.

Any ideas @simskij @rbarry82?
For example, is there a way to ask K8s ahead of time if a given limit is going to be a problem, so I could BlockStatus instead?

rbarry82:
> Problem: when there is a single unit of prometheus and the user sets too high resource limits, then juju is stuck in
> 
> ```
> prometheus-k8s/0  unknown   lost   0/1
> ```
> 
> because the pod cannot be scheduled and there is no charm to take `juju config` changes to repatch the statefulset.
> 
> Any ideas @simskij @rbarry82? For example, is there a way to ask K8s ahead of time if a given limit is going to be a problem, so I could BlockStatus instead?

This is kind of a consistent mess with the kube scheduler. That is -- the kube scheduler is not aware of what else is happening on the system, and a trival process which just `malloc()`s up to 80% will happily let "guaranteed" pods be scheduled which are then OOM killed.

In general, kubelet itself can (and by best practice, does) reserve a certain amount of memory beyond which it will refuse to schedule because the kernel OOM killer may accidentally kill important things (like the kubelet, even, or anything else) otherwise, or swap it out. 

Inside the pod, `/proc` is present. So from a basic level, you can check `/proc/meminfo` and do comparisons to see whether there is free memory. Then, you'd need to determine where the pod is running (if we hope/assume that, once a limit is set, even if the other nodes in the cluster _are_ under `MemoryPressure`, that it can be rescheduled here, though that could also be racy), and describe the node to see whether there's room. Such as:

```
status:
  ...
  allocatable:
    cpu: "4"
    ephemeral-storage: "113180564088"
    hugepages-1Gi: "0"
    hugepages-2Mi: "0"
    hugepages-32Mi: "0"
    hugepages-64Ki: "0"
    memory: 7895304Ki
    pods: "110"
  capacity:
    cpu: "4"
    ephemeral-storage: 122808772Ki
    hugepages-1Gi: "0"
    hugepages-2Mi: "0"
    hugepages-32Mi: "0"
    hugepages-64Ki: "0"
    memory: 7997704Ki
    pods: "110"
```

Lastly, you'd need to check on the pod itself, which may be further limited (if there's no resource limits set in the podspec, this is probably unnecessary paperwork, but regardless) via `/sys/fs/cgroup/memory/memory.limit_in_bytes` inside the pod.

Trivially, checking `/proc/meminfo` to see the amount of free memory (which kube doesn't actually care about, but we may, lest it get OOM killed immediately) versus `status["allocatable"]["memory"]` on a single node under the guess that the cluster is homogenous (and/or the Juju admin deployed with some constraints applied, so we'll end up on an identical worker node) will let you check via differencing whether `/proc/meminfo:MemTotal` and allocatable memory differ to determine the amount of memory reserved in the kubelet (if any -- if there isn't one set, `memory.limit_in_bytes` will be like 9.2 exabytes).

If there isn't a reservation, then all you have to go on is whether a particular node (the one you're running on, maybe) has enough free memory, or you can check all of them, then 🤞 that some _other_ pod which is pending or cannot be scheduled due to `MemoryPressure` doesn't steal it from you when you patch the `StatefulSet`. Or just set `BlockedStatus` and with a message that the requested limits won't be applied.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -0,0 +1,376 @@
...
+        # Make mypy type checking happy that self._patch is a method
+        assert isinstance(self._patch, MethodType)
+        # Ensure this patch is applied during the 'config-changed' event, which is emitted every
+        # startup and every upgrade. The config-changed event is a good time to apply this kind of
+        # patch because it is always emitted after storage-attached, leadership and peer-created,
+        # all of which only fire after install. Patching the statefulset prematurely could result
+        # in those events firing without a workload.
+        self.framework.observe(charm.on.config_changed, self._patch)
+
+        if not refresh_event:
```
balbirthomas, line 217: Can the `refresh_event` argument to the constructor be a "tuple" of `BoundEvent` with a default value of an empty tuple ? If so would this simplify this could logic by eliminating the conditional statement ?

balbirthomas (dismissed):

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -0,0 +1,376 @@
...
+    on = K8sResourcePatchEvents()
+
+    def __init__(
+        self,
+        charm: CharmBase,
+        container_name: str,
+        *,
+        limits: Optional[ResourceSpecDict],
+        requests: Optional[ResourceSpecDict],
+        refresh_event: Optional[Union[BoundEvent, List[BoundEvent]]] = None,
```
Abuelodelanada, line 179: About `refresh_event`... 

Why don't we rename it as `refresh_events` that receives always a `list` no matter if it's one or more than one events?

I think this way the class contract is clearer.... besides the type hinting is cleared too

Abuelodelanada (approved):

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
balbirthomas: Our understanding use cases here has been evolving. So if you think using a non mutable vector default argument makes sense, I see no harm in making Prometheus and Grafana consistent with your choice.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -0,0 +1,490 @@
...
+)
+
+class SomeCharm(CharmBase):
+  def __init__(self, *args):
+    # ...
+    self.resources_patch = KubernetesComputeResourcesPatch(
+        self,
+        "container-name",
+        limits_func=lambda: {"cpu": "1", "mem": "2Gi"},
+        requests_func=lambda: {"cpu": "1", "mem": "2Gi"}
```
Abuelodelanada, line 51: Now that I am integrating this lib into Loki charm I found weird that `KubernetesComputeResourcesPatch` constructor requires a function for `limits_func` and `request_func`.

Why don't we just ask for a dictionary instead?

Abuelodelanada (dismissed):

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
sed-i: It used to be a dict, and recently changed to a func.
Using "dependency injection" made more sense to me because:
- Our convention is to instantiate charm libs in the charm's constructor. The args needed for init are coming raw from what the user set as config options. Processing it in the constructor, obviating, config-changed event, is not easy.
- This way we won't need an "update_X" method we would need to call.
- Bonus: harness currently doesn't support charm reinit between events, and with this approach there are no issue.

Abuelodelanada: mmmm maybe in this case we should revisit our convention about instantiating charm libs in the constructor....

For instance, I was looking the Alertmanager [PR #87](https://github.com/canonical/alertmanager-k8s-operator/pull/87) and, from my point of view, the code is even more complex to understand:

```python
        self.resources_patch = KubernetesComputeResourcesPatch(
            self,
            self._container_name,
            limits_func=lambda: self._resource_limit_from_config(),
            requests_func=lambda: limits_to_requests_scaled(
                self._resource_limit_from_config(),
                ResourceSpecDict(cpu="0.25", memory="200Mi"),
                0.8,
            ),
        )
```

sed-i: Ideally it will be simplified to
```python
        self.resources_patch = KubernetesComputeResourcesPatch(
            self,
            self._container_name,
            limits_func=lambda: self._resource_limit_from_config(),
            requests_func=lambda: ResourceSpecDict(cpu="0.25", memory="200Mi"),
        )
```
if we decide to not scale the requests.

I am happy to discuss alternative designs.

Abuelodelanada: With the last changes this doc section should be updated.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -0,0 +1,562 @@
...
+
+    def is_ready(self) -> bool:
+        """Reports if the resource patch has been applied and is in effect.
+
+        Returns:
+            bool: A boolean indicating if the service patch has been applied and is in effect.
+        """
+        try:
+            resource_reqs = self.resource_reqs_func()
+            limits = resource_reqs.limits
```
Abuelodelanada, line 508: `limits` was removed:


```
  File "/var/lib/juju/agents/unit-loki-0/charm/lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py", line 449, in _patch
    limits = resource_reqs.limits
AttributeError: 'dict' object has no attribute 'limits'
```

Abuelodelanada (changes requested):

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -0,0 +1,562 @@
...
+            self.framework.observe(ev, self._on_config_changed)
+
+    def _on_config_changed(self, _):
+        self._patch()
+
+    def _patch(self) -> None:
+        """Patch the Kubernetes resources created by Juju to limit cpu or mem."""
+        try:
+            resource_reqs = self.resource_reqs_func()
+            limits = resource_reqs.limits
```
Abuelodelanada, line 449: `limits` was removed:

```
  File "/var/lib/juju/agents/unit-loki-0/charm/lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py", line 449, in _patch
    limits = resource_reqs.limits
AttributeError: 'dict' object has no attribute 'limits'
```

Abuelodelanada (dismissed):

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
sed-i: Fixed in next commit.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
sed-i: The `resource_reqs_func` constructor arg is now expected to be `Callable[[], ResourceRequirements]`.
`ResourceRequirements` has `limits` and `requests` attrs.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -129,145 +134,88 @@ def setUp(self, *unused):
...
-    {'cpu': '0.800'}
-    >>> limits_to_requests_scaled({"cpu": "260m"}, {"cpu": "200m"}, sf)
-    {'cpu': '0.208'}
-    >>> limits_to_requests_scaled({"cpu": "250m"}, {"cpu": "200m"}, sf)
-    {'cpu': '0.200'}
-    >>> limits_to_requests_scaled({"cpu": "200m"}, {"cpu": "200m"}, sf)
-    {'cpu': '0.200'}
-    >>> limits_to_requests_scaled({"cpu": "150m"}, {"cpu": "200m"}, sf)
-    {'cpu': '0.150'}
+    TODO: generally, the keys can be anything, not just cpu and memory. Perhaps user could pass
```
simskij, line 226: I think it's totally fine to limit it to only CPU and memory for now.

simskij (dismissed):
Love the reduced complexity by dropping the scaling factor functionality! Good job!

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
sed-i: Lightkube [doesn't restrict](https://github.com/gtsystem/lightkube/pull/37#discussion_r916034989) it so I also opted to not restrict key values. I added [a doctest](https://github.com/canonical/observability-libs/blob/001b3aa6c70704668d5ada96667b2b1bc645a3c7/lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py#L174) for a custom resource.
If you believe we _should_ restrict, I am happy to change this.

sed-i:
Ready for re-review, @simskij @rbarry82 @dstathis.

rbarry82 (approved):

PR #20, opened 2022-06-22
refactored is_valid_uuid method

### LINKED ISSUE #18: Refactor `is_valid_uuid` method
Today the `is_valid_uuid` method in `JujuTopology` class, utilises a regex to validate UUID v4:

```python
def is_valid_uuid(self, uuid):
    """Validates the supplied UUID against the Juju Model UUID pattern."""
    regex = re.compile(
        "^[a-f0-9]{8}-?[a-f0-9]{4}-?4[a-f0-9]{3}-?[89ab][a-f0-9]{3}-?[a-f0-9]{12}$"
    )
    return bool(regex.match(uuid))
```

An old friend told me many years ago:

> _If you have a problem and you use regular expressions to solve it, you now have two problems._

Jokes aside, regular expressions are powerful but complex to read sometimes.

We can avoid regex in this case by using the python `uuid` module. The result is much easy to read and maintain. 

```python
def is_valid_uuid(self, _uuid) -> bool:
     """Validate the supplied UUID against the Juju Model UUID pattern."""
     try:
         return True if uuid.UUID(_uuid).version == 4 else False
     except (TypeError, ValueError):
         return False
```

grequa:
## Issue
We want to avoid using `regex` in the method `is_valid_uuid` to make the code clearer.

Fixes #18.

## Solution
I utilized the module `uuid` instead of `regex` and checked whether the parameter could be passed in as a valid uuid v4.

## Context
By using the `uuid` module we are able to avoid the use of `regex`.

## Testing Instructions
We can test this PR by utilizing the previously implemented tests as this change was a refactor of the previously implemented code. 
`tox -e unit`

## Release Notes
Here, I refactored the `is_valid_uuid` function by implementing the `uuid` module.


.vscode/settings.json
```diff
@@ -0,0 +1,11 @@
+{
```
sed-i, line 1: We usually do not commit IDE config files.
Adding `.vscode` to `.gitignore` might be a good idea though.

sed-i (commented):
Welcome @grequa!
Thanks for the contribution.
A few minor comments below. Also take a look at the [lint results](https://github.com/canonical/observability-libs/runs/7007326057?check_suite_focus=true#step:4:88):
```
lib/charms/observability_libs/v0/juju_topology.py:71:1: F401 're' imported but unused
```

lib/charms/observability_libs/v0/juju_topology.py
```diff
@@ -126,29 +126,19 @@ def __init__(
...
+
+        Args:
+            uuid: string that needs to be checked if it is valid v4 UUID
+        Returns:
+            if parameter is valid v4 UUID
+        """
+        version = 4
+        try:
+            convert_to_uuid = UUID(uuid, version=version)
+            return str(convert_to_uuid) == uuid
```
Abuelodelanada, line 139: I would replace these two lines with only this one:

```python
return True if UUID(uuid).version == 4 else False
```

It's simpler, and besides we do not need to convert the uuid and the compare it with the original one.

Abuelodelanada (changes requested):
Thanks for the PR @grequa !

Please take a look at the test `test_invalid_uuid` in `tests/test_juju_topology.py`. There is a commented case for uuid v4 validation we can uncomment.

lib/charms/observability_libs/v0/juju_topology.py
grequa: We do need to still check if converting the string to uuid changes the value. The code suggested gives an error on `test_invalid_uuid`. To check that my inquiry was correct, I only altered (from the suggested code) the `return True` to be `return str(UUID(uuid))==uuid` therefore adding the additional step to check if the uuid was being altered. This change passed the `test_invalid_uuid`.
The working function would be:
```
def is_valid_uuid(self, uuid):
        
        try:
            return str(UUID(uuid))==uuid if UUID(uuid).version == 4 else False
        except (ValueError, TypeError):
            return False
```
In this case, since we are repetitively calling `UUID(uuid)`, I feel it would be cleaner to write the code the way that I had done so or, better yet, we can write it even more succinctly into one line as follows:
```
def is_valid_uuid(self, uuid):
        
        try:
            return str(UUID(uuid, version=4)) == uuid
        except (ValueError, TypeError):
            return False
```

Abuelodelanada (dismissed):

Abuelodelanada (approved):

sed-i (approved):

PR #21, opened 2022-07-05
promote action

dstathis:
Github action to promote charm taken directly from grafana-agent-k8s.


dstathis:
Script falsely identified this as an actual charm.

PR #22, opened 2022-07-13
Fix: Do not crash when using alternative service name

ghislainbourgeois:
## Issue
When using an alternative service name, Juju first creates a Service with the name of the app. When checking if the Service is already patched, the library crashes because it is trying to read the Service with the alternative name and it does not exists yet, since it was not patched.

## Solution
This change catches the `ApiError` when checking if the Service is patched and returns `False`.

## Context
N/A

## Testing Instructions
Create a charm with a `KubernetesServicePatch` with a custom service name and deploy it successfully.

## Release Notes
Fix: Do not crash when using alternative service name

### DIFF
--- a/lib/charms/observability_libs/v1/kubernetes_service_patch.py
+++ b/lib/charms/observability_libs/v1/kubernetes_service_patch.py
@@ -126,7 +126,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 2
+LIBPATCH = 3
 
 ServiceType = Literal["ClusterIP", "LoadBalancer"]
 
@@ -271,7 +271,15 @@ class KubernetesServicePatch(Object):
 
     def _is_patched(self, client: Client) -> bool:
         # Get the relevant service from the cluster
-        service = client.get(Service, name=self.service_name, namespace=self._namespace)
+        try:
+            service = client.get(Service, name=self.service_name, namespace=self._namespace)
+        except ApiError as e:
+            if e.status.code == 404 and self.service_name != self._app:
+                return False
+            else:
+                logger.error("Kubernetes service get failed: %s", str(e))
+                raise
+
         # Construct a list of expected ports, should the patch be applied
         expected_ports = [(p.port, p.targetPort) for p in self.service.spec.ports]
         # Construct a list in the same manner, using the fetched service
--- a/lib/charms/observability_libs/v1/kubernetes_service_patch.py
+++ b/lib/charms/observability_libs/v1/kubernetes_service_patch.py
@@ -238,10 +238,10 @@ class KubernetesServicePatch(Object):
         except exceptions.ConfigError as e:
             logger.warning("Error creating k8s client: %s", e)
             return
-        if self._is_patched(client):
-            return
 
         try:
+            if self._is_patched(client):
+                return
             if self.service_name != self._app:
                 self._delete_and_create_service(client)
             client.patch(Service, self.service_name, self.service, patch_type=PatchType.MERGE)
--- a/tests/test_kubernetes_service.py
+++ b/tests/test_kubernetes_service.py
@@ -453,3 +453,41 @@ class TestK8sServicePatch(unittest.TestCase):
         self.assertEqual(self.harness.charm.service_patch._app, "test-charm")
         self.assertEqual(self.harness.charm.service_patch._namespace, "test")
         mock.assert_called_with("/var/run/secrets/kubernetes.io/serviceaccount/namespace", "r")
+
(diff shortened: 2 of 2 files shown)


rbarry82 (approved):
Looks ok to me in general, but do we have information about when this service will actually be created? What's the original producer of this bug?

lib/charms/observability_libs/v1/kubernetes_service_patch.py
```diff
@@ -271,7 +271,11 @@ def is_patched(self) -> bool:
 
     def _is_patched(self, client: Client) -> bool:
         # Get the relevant service from the cluster
-        service = client.get(Service, name=self.service_name, namespace=self._namespace)
+        try:
+            service = client.get(Service, name=self.service_name, namespace=self._namespace)
```
sed-i, line 275: ApiError seems a bit too broad to conclude with `False` ([example](https://github.com/canonical/observability-libs/blob/4e20f6ceaa7d51cd1b32aa6eb313fcc7ce371cec/lib/charms/observability_libs/v1/kubernetes_service_patch.py#L248)). Would it make sense to first use `client.list()` to check if the service exists?

sed-i (commented):
Thanks @ghislainbourgeois!
Bump LIBPATCH?

lib/charms/observability_libs/v1/kubernetes_service_patch.py
ghislainbourgeois: Good point, I have changed the code a bit to only return `False` when the status is `404` and that we are using a custom name. For all the other cases, I log an error and raise. (Not raising there would crash on the next instruction)

sed-i: Thanks!

The approach we try to take with charm libs is to never to raise, but instead emit a custom event, if necessary (this could be a separate issue if needed). This lib too logs and returns instead of raising ([example](https://github.com/canonical/observability-libs/blob/4e20f6ceaa7d51cd1b32aa6eb313fcc7ce371cec/lib/charms/observability_libs/v1/kubernetes_service_patch.py#L240)).

Is there a particular reason to raise here?

tests/test_kubernetes_service.py
```diff
@@ -453,3 +453,41 @@ def test_patch_properties(self, mock):
...
+        self.harness.set_leader(False)
+
+        client.return_value = client
+        client.get.side_effect = _FakeApiError(404)
+        self.assertEqual(charm.custom_service_name_service_patch.is_patched(), False)
+
+    @patch(f"{MOD_PATH}.ApiError", _FakeApiError)
+    @patch(f"{CL_PATH}._namespace", "test")
+    @patch(f"{MOD_PATH}.Client")
+    def test_is_patched_k8s_service_api_error_without_alternative_name(self, client):
```
sed-i, line 471: Would this improve clarity?

```suggestion
    def test_is_patched_k8s_service_api_error_with_default_name(self, client):
```

lib/charms/observability_libs/v1/kubernetes_service_patch.py
ghislainbourgeois: The main reason is that the code will crash anyway with another exception if I let it go further because `service` will be `None`. We can move the rest of the method inside the `try` block, but then we will need to return something in the `except` block. It is the main reason I initially returned `False` on all errors. We could however catch the exception up the stack and just log the error.

tests/test_kubernetes_service.py
ghislainbourgeois: Yes, thanks!

lib/charms/observability_libs/v1/kubernetes_service_patch.py
ghislainbourgeois: @sed-i Any preference on returning `False` or raising? I do not feel like any of the solutions are perfect, but either way will fix the crash that we are experiencing in some charms.

sed-i: Since we have a public interface,
```python
def is_patched(self) -> bool:
```
I think raising is preferred.

However, when called internally from within the module as part of `def _patch()`, I think we need to absorb the exception to remain consistent:
```diff
-if self._is_patched(client):
-    return

try:
+   if self._is_patched(client):
+       return
    if self.service_name != self._app:
        self._delete_and_create_service(client)
```

ghislainbourgeois: Makes sense, I applied the change.

sed-i (approved):

sed-i:
LG2M, thanks!

jnsgruk (approved):

jnsgruk:
Merged and published to the store :)

PR #32, opened 2022-10-14
Gracefully catch SIGTERMs in KubernetesComputeResourcesPatch

rbarry82:
## Issue
When the bundle is deployed, with resource limits applied, we are seeing occasional races which result in "random" hooks failing ([`grafana-dashboard-relation-created`](https://github.com/canonical/grafana-k8s-operator/issues/140), @sed-i has seen `leader-elected` and `peer-relation-joined` fail in integration tests, etc). Juju's `caasapplicationterminator` sends a `SIGTERM`, nicely, but without a signal handler, Python bails and we see `hook-failed: foo-bar-baz`

## Solution
If the patch is not already applied, attach a `SIGTERM` handler prior to patching. 

## Context
In case of `SIGTERM` (a tombstone from Juju, really), Juju will re-try the hook, so as long as the charms stay idempotent, there's no risk here either.

## Testing Instructions
It's a race, so unfortunately, it's kind of "fetch the patch and repeatedly try manually until we're reasonably confident" without synthetic integration tests which just sleep with a handler, which don't really prove much other than "Juju behaves" as far as the actual race, and will run up the integration test time without a "real" workflow.

## Release Notes
Gracefully catch SIGTERMs in KubernetesComputeResourcesPatch

### DIFF
--- a/lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
+++ b/lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
@@ -103,6 +103,8 @@ References:
 
 import decimal
 import logging
+import signal
+import sys
 from decimal import Decimal
 from math import ceil, floor
 from typing import Callable, Dict, List, Optional, Union
@@ -133,7 +135,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
+LIBPATCH = 2
 
 
 _Decimal = Union[Decimal, float, str, int]  # types that are potentially convertible to Decimal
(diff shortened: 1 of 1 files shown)


lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -399,13 +401,28 @@ def is_ready(self, pod_name, resource_reqs: ResourceRequirements):
             resource_reqs, self.get_actual(pod_name)
         )
 
+    def _handle_pod_termination(self, *args) -> None:
+        logger.debug(
+            "KubernetesComputeResourcesPatch was interrupted by a SIGTERM, likely due to "
+            "pod termination during execution of `config-changed`. Exiting gracefully. "
+            "The hook will be re-run"
```
PietroPasotti, line 408: Did you forget to finish the sentence or is it a metaphor for the fact that the charm was killed in flight?
Perhaps it would be helpful to clarify when we _think_ this re-execution will take place?
"The hook that was being ran right now will be re-run by juju as soon as the pod is respawned."

PietroPasotti (dismissed):
Curious to see if this will work. It does look like a low-risk patch either way, so we might just merge and see if it does it.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -399,13 +401,28 @@ def is_ready(self, pod_name, resource_reqs: ResourceRequirements):
             resource_reqs, self.get_actual(pod_name)
         )
 
+    def _handle_pod_termination(self, *args) -> None:
+        logger.debug(
+            "KubernetesComputeResourcesPatch was interrupted by a SIGTERM, likely due to "
```
sed-i, line 406:
```suggestion
            "KubernetesComputeResourcesPatch's signal handler was triggered by a SIGTERM, likely due to "
```

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -399,13 +401,28 @@ def is_ready(self, pod_name, resource_reqs: ResourceRequirements):
...
         """Patch the Kubernetes resources created by Juju to limit cpu or mem."""
         # Need to ignore invalid input, otherwise the StatefulSet gives "FailedCreate" and the
         # charm would be stuck in unknown/lost.
         if self.is_patched(resource_reqs):
             return
 
+        # If it's not patched already, add a handler for SIGTERM prior to patching.
+        # Juju tries to send a SIGTERM to the CRI to exit gracefully when in CAAS mode, then
+        # the hook is re-executed, so we can "safely" trap it here without causing a hook
+        # failure if there is a race, and config-changed will retry (after it is applied and
```
sed-i, line 422:
```suggestion
        # failure if there is a race, and config-changed will retry (after it is applied) and
```

sed-i (approved):

PR #37, opened 2023-02-01
Pipify 

sed-i:
## Issue
In tandem with:
- https://github.com/canonical/prometheus-k8s-operator/pull/444
- https://github.com/canonical/avalanche-k8s-operator/pull/22


dstathis:
`JujuTopology` seems to be implemented in `__init__.py` and `juju_topology.py`. It should be only in `juju_topology.py` and imported in `__init__.py`

dstathis:
This one is not a deal breaker by any means but I thought I would mention it anyway. I tend to not be a fan of names that have numbers in them like `jo11y` because they are annoying to type.

PietroPasotti:
> What is this empty file `py.typed`?
[cfr](https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/)

dstathis:
We should probably try to use a standard python library layout with the lib name as the source dir. So it would be something like 
```
- README.md
- pyproject.toml
...
- jo11y
  - __init__.py
  - juju_topology.py
  ...
```

Also, this is on it's own branch so I would just put the files in `placeholder` where they properly belong so we can look at the resulting repo as it would look.

sed-i:
> `JujuTopology` seems to be implemented in `__init__.py` and `juju_topology.py`. It should be only in `juju_topology.py` and imported in `__init__.py`

Not sure what you mean @dstathis 
- `JujuTopology` is implemented in `juju_topology.py` and `__init__.py` is an empty file, as prescribed.
- As a result, the import statement is `from jo11y.juju_topology import JujuTopology`. What would you like to have?



> I tend to not be a fan of names that have numbers in them like `jo11y` because they are annoying to type.

Yes, `jo11y` is annoying to type but it seems to me that the jolly + o11y reference is worth it. Wdyt?



> We should probably try to use a standard python library layout

Agreed. I followed an [official tutorial](https://packaging.python.org/en/latest/tutorials/packaging-projects/). I guess that's the recommended layout with `hatchling`.

sed-i:
CI is green and I think the experiment is complete.
After using our utils as a pypi pkg I agree it's more ergonomic for users and devs than `fetch-lib charms.observability_libs.v0.jolly` and `from charms.observability_libs.v0.jolly import JujuTopology, CosTool`.

Regarding polluting pypi with tightly coupled to ops, I'd be happy if we name it `ops_jo11y` or `ops[jo11y]`.

rbarry82 (dismissed):

sed-i:
Any idea @rbarry82 what could have changed between [639821e](https://github.com/canonical/observability-libs/pull/37/commits/639821e3e706fc663e7aae508145661f790b7bcc) (which passed CI) and now?
(I'm having trouble with running the itests on the mac.)

rbarry82:
I also can't run them on Linux because Lightkube expects to find a token that isn't there on this repo for some reason. No real changes that pop out at me, but I loosened the requirement checking (so at least it gets python-libjuju 3.0.4 instead of 3.0.0). When I'm on the other side of the Atlantic, I'll have more time to investigate.

rbarry82 (approved):

PR #38, opened 2023-02-06
Update endpoint discovery

rbarry82:
## Issue
`juju-run` is no longer present in juju3, replaced with `juju-exec`. Do it here also.

Start the observer as part of the constructor, and squash the payload so it doesn't get duplicate records sent.

### DIFF
--- a/lib/charms/observability_libs/v0/metrics_endpoint_discovery.py
+++ b/lib/charms/observability_libs/v0/metrics_endpoint_discovery.py
@@ -45,6 +45,7 @@ import os
 import signal
 import subprocess
 import sys
+from pathlib import Path
 from typing import Dict, Iterable
 
 from lightkube import Client
@@ -62,7 +63,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 2
+LIBPATCH = 3
 
 # File path where metrics endpoint change data is written for exchange
 # between the discovery process and the materialised event.
@@ -126,12 +127,11 @@ class MetricsEndpointObserver(Object):
         self._observer_pid = 0
 
         self._labels = labels
+        self.start_observer()
 
     def start_observer(self):
         """Start the metrics endpoint observer running in a new process."""
         self.stop_observer()
-
-        logging.info("Starting metrics endpoint observer process")
 
         # We need to trick Juju into thinking that we are not running
         # in a hook context, as Juju will disallow use of juju-run.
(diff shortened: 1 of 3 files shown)


lib/charms/observability_libs/v0/metrics_endpoint_discovery.py
```diff
@@ -126,35 +127,40 @@ def __init__(self, charm: CharmBase, labels: Dict[str, Iterable]):
...
 
-        logging.info("Starting metrics endpoint observer process")
-
         # We need to trick Juju into thinking that we are not running
         # in a hook context, as Juju will disallow use of juju-run.
         new_env = os.environ.copy()
         if "JUJU_CONTEXT_ID" in new_env:
             new_env.pop("JUJU_CONTEXT_ID")
 
+        tool_prefix = f"/var/lib/juju/tools/{self.unit_tag}"
```
PietroPasotti, line 142: I thought we didn't need to support juju2. Do we?

lib/charms/observability_libs/v0/metrics_endpoint_discovery.py
```diff
@@ -126,35 +127,40 @@ def __init__(self, charm: CharmBase, labels: Dict[str, Iterable]):
         self._observer_pid = 0
 
         self._labels = labels
+        self.start_observer()
 
     def start_observer(self):
```
PietroPasotti, line 132: should this become private, now that we're calling it from init? Or is there still a reason/way to call this from somewhere else (external)?

PietroPasotti (commented):
Not quite aware of the background of this change, so just two flyby comments from me :)

lib/charms/observability_libs/v0/metrics_endpoint_discovery.py
rbarry82: _We_ do not, but KubeFlow is staying on 2.9 because it's LTS 🤷

lib/charms/observability_libs/v0/metrics_endpoint_discovery.py
rbarry82: Since it was previously visible and there's not real data mutation from leaving public, I didn't see a need to make it private. I don't actually think anyone is using it currently (or the public method), but still...

lib/charms/observability_libs/v0/metrics_endpoint_discovery.py
rbarry82: We do point to `/usr/bin/juju-exec` in this patch if the other doesn't exist. I don't have a 2.9 environment to test against here, and the assumption was that the path in [current version of the library](https://github.com/canonical/observability-libs/blob/main/lib/charms/observability_libs/v0/metrics_endpoint_discovery.py#L147) was correct. If you can verify that it actually is `/usr/bin/juju-run` on 2.9, we should also change that.

sed-i: Confirmed :point_up: with 2.9.38.

rbarry82: Done.

PR #39, opened 2023-02-07
check version bump on library modifications and fix linting

lucabello:
## Issue
Forgetting to bump `LIBPATCH` on a library update causes that update to not be detected at any level, as the CI won't release a new library, but it can be detected by the PR label, leading to confusion.

## Solution
Add a simple check to the `static-lib` environment to make sure that *if* a library has been modified, its `LIBPATCH` has been bumped. Once this is approved and merged, I will replicate it to all of our other repos.   
I'm open to have this somewhere else if you think it's better, such as its own `tox` environment (which would require modifying the *_quality_checks* CI to also run that step), or anywhere else.

## Context
Note that this is intentionally a loose check (it only checks if the `LIBPATCH` line was modified); this allows for *some* flexibility in not bumping the version by just adding a comment to that line.  
However, if there's a need for multiple PRs on the same library, and we don't want to release the library every time, the correct approach is probably to create another branch out of main (i.e., `library-change`); then merge the different PRs to `library-change`, and finally open a PR from that branch to main.

### DIFF
--- a/tests/unit/test_kubernetes_compute_resources.py
+++ b/tests/unit/test_kubernetes_compute_resources.py
@@ -74,7 +74,7 @@ class TestKubernetesComputeResourcesPatch(unittest.TestCase):
         self.harness.begin_with_initial_hooks()
 
         # Test invalid quantity values
-        for (cpu, memory) in [
+        for cpu, memory in [
             ("-1", ""),
             ("", "-1Gi"),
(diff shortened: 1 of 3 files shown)


lucabello:
I'm also adding a commit to fix the linting after its update

rbarry82 (approved):

PR #40, opened 2023-02-07
add badges to README.md

lucabello:


README.md
```diff
@@ -1,5 +1,10 @@
 # Observability Libs
 
+[![CharmHub Badge](https://charmhub.io/observability-libs/badge.svg)](https://charmhub.io/observability-libs)
```
sed-i, line 3: This is a charm badge, and this charm is a tester/dummy one.
We probably can skip the charm badge for utils/relation repos.

sed-i (approved):

README.md
lucabello: True, removing that

sed-i (approved):

PR #42, opened 2023-03-01
Remove the DeprecationWarning

rbarry82:
## Issue
We're not releasing for a bit while we work on other things, and the `DeprecationWarning` flags almost every time this is instantiated. 

Get rid of it until we're ready.


jnsgruk (approved):
Good call :)

PR #45, opened 2023-03-28
simplify logic, remove stored state dep, fix mem leak

simskij:
## Issue

https://github.com/canonical/seldon-core-operator/issues/123 
Every event processed by the charm lead to another process instance.

## Solution

Fixes the issue by removing the need for pid juggling and instead doing a blanket `pkill` for everything that matches our invocation. 

## Context

N/A

## Testing Instructions

1. Copy lib to seldon-core
2. Deploy seldon-core
3. Check `ps ax` for duplicate processes

### DIFF
--- a/lib/charms/observability_libs/v0/metrics_endpoint_discovery.py
+++ b/lib/charms/observability_libs/v0/metrics_endpoint_discovery.py
@@ -62,7 +62,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 3
+LIBPATCH = 4
 
 # File path where metrics endpoint change data is written for exchange
 # between the discovery process and the materialised event.
--- a/lib/charms/observability_libs/v0/metrics_endpoint_discovery.py
+++ b/lib/charms/observability_libs/v0/metrics_endpoint_discovery.py
@@ -162,7 +162,7 @@ class MetricsEndpointObserver(Object):
     def stop_observers(self):
         """Stops all running instances of the observer."""
         logging.info("Stopping all running metrics endpoint observer processes")
-        subprocess.run(["pkill", "--signal", "INT", "-f", ".*metrics_endpoint.*"])
+        subprocess.run(["pkill", "--signal", "INT", "-f", ".*lib/charms/observability_libs/v[0-9]/metrics_endpoint_discovery\.py.*"])
 
     @property
     def unit_tag(self):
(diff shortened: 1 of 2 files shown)


PietroPasotti (commented):
How high is the risk that someone else has created another process containing `metrics_endpoint.`?

Do we risk killing something we don't own?
At least we need to make this very clear in the docs, and perhaps log with debug level a full list of everything we're killing so if someone ever wonders 'where did my obtusely named process go'? they have a lead

simskij:
> How high is the risk that someone else has created another process containing `metrics_endpoint.`?
> 
> Do we risk killing something we don't own? At least we need to make this very clear in the docs, and perhaps log with debug level a full list of everything we're killing so if someone ever wonders 'where did my obtusely named process go'? they have a lead

I can make it more targeted. 👍🏼

rbarry82:
This is fine with me in general, but it would be interesting to know _why_ selecting the process via a stored PID is not working.

`pkill` is not particularly portable (older alpines, `ubi[-minimal]`, and some others don't have it at all, and we know that there's a large-ish mix in some of the containers used), and using `shell=True` is better avoided.

simskij:
> This is fine with me in general, but it would be interesting to know _why_ selecting the process via a stored PID is not working.
> 
> `pkill` is not particularly portable (older alpines, `ubi[-minimal]`, and some others don't have it at all, and we know that there's a large-ish mix in some of the containers used), and using `shell=True` is better avoided.

I agree with you in principle. But I also feel this lib is low-value and makeshift enough to not warrant any deeper investigation.

dstathis (approved):

simskij:
> This is fine with me in general, but it would be interesting to know _why_ selecting the process via a stored PID is not working.
> 
> `pkill` is not particularly portable (older alpines, `ubi[-minimal]`, and some others don't have it at all, and we know that there's a large-ish mix in some of the containers used), and using `shell=True` is better avoided.



> > This is fine with me in general, but it would be interesting to know _why_ selecting the process via a stored PID is not working.
> > `pkill` is not particularly portable (older alpines, `ubi[-minimal]`, and some others don't have it at all, and we know that there's a large-ish mix in some of the containers used), and using `shell=True` is better avoided.
> 
> I agree with you in principle. But I also feel this lib is low-value and makeshift enough to not warrant any deeper investigation.

Also, this is in the charm container, so that will always be Ubuntu. 👍🏼

PietroPasotti (approved):

rbarry82 (approved):

PR #49, opened 2023-06-13
Add cert_handler library

lucabello:
## Issue
This PR addresses the issue of implementing TLS between the components of the **cos-lite** bundle.

## Solution
The **cert_manager.py** library is a convenient wrapper around the `tls_certificates` library. It allows integrating TLS in charms quite smoothly.

## Testing Instructions
Use this library to obtain a certificate in a charm and try to use it.


lucabello:
Currently the main point of contention is if the `cert_manager.py` library should have a lib dependency on `tls_certificates` or not.

It has been mentioned that since the dependency is only on our charms side, an import guard could suffice as the final implementation would be cleaner. This is the approach that's currently implemented in this PR.

Illustrating both scenarios in terms of charm integration:
* **no library dependency**: the charm needs to pass all the relevant `tls_certificates` functions to `CertManager` and observe its custom events on its own; the charm would contain something like this:
  
  ```python
  import charms.tls_certificates_interface.v2.tls_certificate
  import charms.observability_libs.v0.cert_manager
  
  self.certificates = tls_certificates.TLSCertificatesRequiresV2(self, "certificates")
  
  self.cert_manager = CertManager(self,
    private_key_password=b"password",
    generate_private_key=tls_certificates.generate_key,
    generate_csr=tls_certificates.generate_csr,
    request_certificate_creation=tls_certificates.request_certificate_creation,
    request_certificate_renewal=tls_certificates.request_certificate_renewal
  )
  
  self.framework.observe(self.on.certificates_relation_joined, self.cert_manager._on_certificates_relation_joined)
  self.framework.observe(self.certificates.on.certificate_available, self.cert_manager._on_certificate_available)
  self.framework.observe(self.certificates.on.certificate_expiring, self.cert_manager._on_certificate_expiring)
  self.framework.observe(self.certificates.on.certificate_invalidated, self.cert_manager._on_certificate_invalidated)
  self.framework.observe(self.certificates.on.all_certificates_invalidated, self.cert_manager._on_all_certificates_invalidated)
  ```
  
* **with library dependency**: this scenario is much simpler as the charm can delegate everything to the library:
  
  ```python
  import charms.observability_libs.v0.cert_manager
  
  self.cert_manager(self,
    private_key_password=b"password"
  )
  ```

lucabello:
Another useful feature to include in this library, I think, is to include saving the certificate to file once it's obtained, instead of only having it in peer relation (possibly by passing a folder path to the `CertManager` object itself, or having a good default).

This way a charm only needs to worry about using the certificate.

What do you think about it?

lib/charms/observability_libs/v0/cert_handler.py
```diff
@@ -0,0 +1,343 @@
...
+    )
+from ops.charm import CharmBase, RelationBrokenEvent
+from ops.model import Relation
+from ops.framework import EventBase, EventSource, Object, ObjectEvents
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+LIBID = "deadbeef"
```
PietroPasotti, line 54: we probably need to publish it and get a libid before we merge

sed-i: Done.

lib/charms/observability_libs/v0/cert_handler.py
```diff
@@ -0,0 +1,361 @@
...
+
+    CerHandler manages one single cert.
+
+    Args:
+        key: A manually-crafted, static, unique identifier used by ops to identify events.
+        It shouldn't change between one event to another.
+        peer_relation_name: Must match metadata.yaml.
+        certificates_relation_name: Must match metadata.yaml.
+        cert_subject: Custom subject. Name collisions are under the caller's responsibility.
+        extra_sans_dns: Any additional DNS names apart from FQDN.
```
Abuelodelanada, line 81: We usually place these arguments in the `__init__` docstring.

Abuelodelanada (dismissed):

lib/charms/observability_libs/v0/cert_handler.py
sed-i: Fixed in next commit.

lib/charms/observability_libs/v0/cert_handler.py
```diff
@@ -0,0 +1,343 @@
...
+following example:
+
+```python
+cert_handler = CertHandler(
+    charm=self,
+    peer_relation_name="replicas",
+    cert_subject="unit_name",  # Optional
+    key="cert-manager"  # Optional
+)
+```
```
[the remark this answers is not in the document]
sed-i: Fixed in next commit.

sed-i (approved):

PR #51, opened 2023-07-13
Add extra useful methods and one bugfix.

dstathis:
Adds some useful methods and fixes a bug where we assumed the relation name was "certificates".


lucabello (dismissed):
This implies that in the timeframe between relating a charm to a certificate provider and actually getting the certificate, we don't want the charm to keep working over HTTP, and we'd rather have it fail instead. 

I'd wait to merge this until we hear what people think about it, but I'm approving for a fast merge in case we agree on this :)

lucabello (approved):

PR #60, opened 2023-09-19
Split SANs into DNS and IP

sed-i:
## Issue
When traefik generates a CSR with IP address (this is often the case in testing), then the IP address goes into the DNS section, but according to [RFC 2818](https://www.rfc-editor.org/rfc/rfc2818) ([via](https://stackoverflow.com/a/9200012/3516684)) it needs to go to a different, dedicated section:

>   In some cases, the URI is specified as an IP address rather than a
>   hostname. In this case, the iPAddress subjectAltName must be present
>   in the certificate and must exactly match the IP in the URI.

## Solution
Keep cert_handler API the same, but internally split the list into ips and hostnames and pass accordingly to `generate_csr`.

Tandem PRs:
- https://github.com/canonical/traefik-k8s-operator/pull/245
- https://github.com/canonical/cos-lite-bundle/pull/80
- https://github.com/canonical/traefik-k8s-operator/pull/249

Possibly depends on:
- https://github.com/canonical/tls-certificates-interface/issues/71

## Context
NTA.

## Testing Instructions
Deploy the tandem PRs. Then copy over the external-ca cert, and attempt to curl prometheus via its ingress url, from within the grafana container.

```bash
juju run external-ca/0 get-ca-certificate --format json | jq -r '."external-ca/0".results."ca-certificate"' > external-ca.crt
openssl x509 -noout -text -in external-ca.crt

echo | openssl s_client -showcerts -servername 10.43.8.206 -connect 10.43.8.206:443 | openssl x509 -text -noout

juju scp --container prometheus external-ca.crt prometheus/0:/usr/local/share/ca-certificates
juju ssh --container prometheus prometheus/0 update-ca-certificates --fresh

juju scp --container grafana external-ca.crt grafana/0:/usr/local/share/ca-certificates
juju ssh --container grafana grafana/0 update-ca-certificates --fresh

juju ssh --container grafana grafana/0 curl https://10.43.8.206/trfk-prometheus-0/api/v1/targets | jq | grep -E "scrapeUrl|health"

# from within the grafana workload container, from /usr/local/share/ca-certificates
openssl verify -CAfile $PWD/external-ca.crt -CApath $PWD -verbose -x509_strict -verify_ip ip
```

Before https://github.com/canonical/traefik-k8s-operator/pull/249, curl gives `curl: (60) SSL certificate problem: self-signed certificate` and `openssl verify` complains:
```
error 85 at 0 depth lookup: Missing Authority Key Identifier
C = US, CN = external.demo.ca
error 92 at 1 depth lookup: CA cert does not include key usage extension
error stdin: verification failed
```

With the default cert set in traefik, it works.

## Release Notes
Split SANs into DNS and IP.

### DIFF
--- a/lib/charms/observability_libs/v0/cert_handler.py
+++ b/lib/charms/observability_libs/v0/cert_handler.py
@@ -33,8 +33,10 @@ container.push(certpath, self.cert_handler.cert)
 This library requires a peer relation to be declared in the requirer's metadata. Peer relation data
 is used for "persistent storage" of the private key and certs.
 """
+import ipaddress
 import json
 import socket
+from itertools import filterfalse
 from typing import List, Optional, Union
 
 try:
@@ -62,7 +64,16 @@ logger = logging.getLogger(__name__)
 
 LIBID = "b5cd5cd580f3428fa5f59a8876dcbe6a"
 LIBAPI = 0
-LIBPATCH = 7
+LIBPATCH = 8
+
+
+def is_ip_address(value: str) -> bool:
+    """Return True if the input value is a valid IPv4 address; False otherwise."""
+    try:
+        ipaddress.IPv4Address(value)
+        return True
+    except ipaddress.AddressValueError:
+        return False
 
 
 class CertChanged(EventBase):
@@ -88,7 +99,7 @@ class CertHandler(Object):
         peer_relation_name: str,
         certificates_relation_name: str = "certificates",
         cert_subject: Optional[str] = None,
-        extra_sans_dns: Optional[List[str]] = None,
+        extra_sans_dns: Optional[List[str]] = None,  # TODO: in v1, rename arg to `sans`
     ):
         """CertHandler is used to wrap TLS Certificates management operations for charms.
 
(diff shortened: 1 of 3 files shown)


sed-i:
@ghislainbourgeois have you tried using CSRs that list bare IPs? Did it work for you?

sed-i:
According to [this](https://security.stackexchange.com/a/169775/58740), we need to add [`x509.KeyUsage`](https://cryptography.io/en/latest/x509/reference/#cryptography.x509.KeyUsage)  to [`generate_ca`](https://github.com/canonical/tls-certificates-interface/blob/d08dc35426ae6a790a3f6fc11088f29840ceb217/lib/charms/tls_certificates_interface/v2/tls_certificates.py#L652).

[Example usage](https://stackoverflow.com/a/66360937/3516684).

Wdyt @ghislainbourgeois?

sed-i:
Confirmed this works with https://github.com/canonical/traefik-k8s-operator/pull/249.

lib/charms/observability_libs/v0/cert_handler.py
```diff
@@ -111,7 +122,9 @@ def __init__(
         self.cert_subject = charm.unit.name.replace("/", "-") if not cert_subject else cert_subject
 
         # Use fqdn only if no SANs were given, and drop empty/duplicate SANs
-        self.sans_dns = list(set(filter(None, (extra_sans_dns or [socket.getfqdn()]))))
+        sans = list(set(filter(None, (extra_sans_dns or [socket.getfqdn()]))))
```
PietroPasotti, line 125: minor code readability nit:

```suggestion
        if extra_sans_dns:
            sans = filter(None, extra_sans_dns)
        else:
            sans = [socket.getfqdn()]
        
        self.sans_ips = sans_ips = []
        self.sans_dns = sans_dns = []
        
        for name in sans:
            if is_ip_address(name):
                sans_ips.append(name)
            else:
                sans_dns.append(name)
```

PietroPasotti (approved):
only a stylistic nit -- that code felt hard to read and understand

lib/charms/observability_libs/v0/cert_handler.py
sed-i: I see what you mean.
I'm on the fence on this one because:
- filtering with a predicate is a well established construct that conveys intention effectively
- 3 no-so-long lines vs. 12 lines

PR #61, opened 2023-10-11
upkeep: use refactored ci and pyright

lucabello:


lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -322,7 +322,7 @@ def __init__(self, namespace: str, statefulset_name: str, container_name: str):
         self.namespace = namespace
         self.statefulset_name = statefulset_name
         self.container_name = container_name
-        self.client = Client()
+        self.client = Client()  # pyright: ignore
```
PietroPasotti, line 325: can we solve whatever issue it is forcing us to add this ignore instead?

PietroPasotti (approved):

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
PietroPasotti: mm, it seems to be in the upstream

lib/charms/observability_libs/v0/metrics_endpoint_discovery.py
```diff
@@ -47,7 +47,7 @@ def _on_endpoints_change(self, event):
 from pathlib import Path
 from typing import Dict, Iterable
 
-from lightkube import Client
+from lightkube import Client  # pyright: ignore
```
sed-i, line 50: Iirc, if we import `from lightkube.client` we wouldn't need the ignore.

src/charm.py
```diff
@@ -45,7 +45,7 @@ def _resource_spec_from_config(self) -> ResourceRequirements:
         return adjust_resource_requirements(resource_limit, None)
 
     def _on_resource_patch_failed(self, event: K8sResourcePatchFailedEvent):
-        self.unit.status = BlockedStatus(event.message)
+        self.unit.status = BlockedStatus(str(event.message))
```
sed-i, line 48: We shouldn't need a `str` here. We should probably fix the type of `message`.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
PietroPasotti: Client's signature has something like `foo: Bar = None` and pyright seems unhappy with that 'implicit' `Optional`.
We could submit a pr to the upstream to make it `foo: Optional[Bar] = None`, but I'm not sure if it's worth the effort, and if 'being compatible with pyright' is a good reason for a third-party product to accept a change.

lucabello: Correct about the issue being upstream; this is something we already did for everything else importing lightkube :)

lib/charms/observability_libs/v0/metrics_endpoint_discovery.py
lucabello: Yes and no, this ends up being a lot easier; as a reminder, `mypy` was happily skipping this until now :)

src/charm.py
lucabello: This charm is not even published, so I doubt we would notice the error from this charm; regardless it's a pretty small change, so sure!

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
sed-i: What is the issue? Can you paste the error pyright prints?

lib/charms/observability_libs/v0/cert_handler.py
```diff
@@ -363,7 +363,7 @@ def _server_cert(self, value: str):
     def _chain(self) -> List[str]:
         if self._peer_relation:
             if chain := self._peer_relation.data[self.charm.unit].get("chain", []):
-                return json.loads(chain)
+                return json.loads(str(chain))
```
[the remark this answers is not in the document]
lucabello: Done!

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
lucabello:
```
/home/aegis/Repositories/Canonical/observability-libs/lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py:110:33 - error: "Client" is not exported from module "lightkube"
```

Pietro described the issue with it already; again, this *could* be fixed by importing the `Client` directly, but it's a lot of code changes, and I'd rather keep this PR small; just as a reminder, `mypy` wasn't checking any of this, that's why we didn't get errors :)

PR #66, opened 2023-11-21
Move private key from peer data to juju secrets

Abuelodelanada:
## Issue

This PR adds a new cert_handler version that moves key and certs from peer relation to Juju secrets


tests/scenario/test_cert_handler/test_cert_handler_v1.py
```diff
@@ -0,0 +1,42 @@
...
+    return Relation("certificates")
+
+
+@pytest.mark.parametrize("leader", (True, False))
+def test_cert_joins(ctx, certificates, leader):
+    with ctx.manager(
+        certificates.joined_event, State(leader=leader, relations=[certificates])
+    ) as runner:
+        runner.run()
+        assert runner.charm.ch._private_key
```
PietroPasotti, line 42: the tests are a little bit skinny IMHO. We're major-bumping a lib that all our charms are using, we should take the opportunity to add some finer-grained tests.

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -0,0 +1,391 @@
...
+
+        self.certificates_relation_name = certificates_relation_name
+        self.certificates = TLSCertificatesRequiresV2(self.charm, self.certificates_relation_name)
+
+        self.framework.observe(
+            self.charm.on.config_changed,
+            self._on_config_changed,
+        )
+        self.framework.observe(
+            self.charm.on.certificates_relation_joined,  # pyright: ignore
```
PietroPasotti, line 137: shouldn't this be `self.charm.on[self.certificates_relation_name].relation_joined`?

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -0,0 +1,391 @@
...
+            self._on_all_certificates_invalidated,
+        )
+        self.framework.observe(
+            self.charm.on[self.certificates_relation_name].relation_broken,  # pyright: ignore
+            self._on_certificates_relation_broken,
+        )
+
+    @property
+    def enabled(self) -> bool:
+        """Boolean indicating whether the charm has a tls_certificates relation."""
```
PietroPasotti, line 163: ... and that the relation is healthy (there is a remote app, there is a databag...) to filter out some common 'relation-breaking' data model inconsistencies.

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -0,0 +1,391 @@
...
+            len(self.charm.model.relations[self.certificates_relation_name]) > 0
+            and len(self.charm.model.get_relation(self.certificates_relation_name).units) > 0  # type: ignore
+        )
+
+    def _on_certificates_relation_joined(self, _) -> None:
+        """Generate the CSR and request the certificate creation."""
+        self._generate_privkey()
+        self._generate_csr()
+
+    def _on_certificates_relation_changed(self, _) -> None:
```
PietroPasotti, line 176: should we delete this? or is it meant to be overridden?

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -0,0 +1,391 @@
...
+
+You can then observe the library's custom event and make use of the key and cert:
+```python
+self.framework.observe(self.cert_handler.on.cert_changed, self._on_server_cert_changed)
+
+container.push(keypath, self.cert_handler.key)
+container.push(certpath, self.cert_handler.cert)
+```
+
+This library requires a peer relation to be declared in the requirer's metadata. Peer relation data
```
PietroPasotti, line 32: this can go

PietroPasotti: and maybe we should add a notice that this library can't be used if your juju version doesn't support secrets?

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -0,0 +1,391 @@
...
+        Args:
+            charm: The owning charm.
+            key: A manually-crafted, static, unique identifier used by ops to identify events.
+             It shouldn't change between one event to another.
+            certificates_relation_name: Must match metadata.yaml.
+            cert_subject: Custom subject. Name collisions are under the caller's responsibility.
+            sans: DNS names. If none are given, use FQDN.
+        """
+        super().__init__(charm, key)
+
```
PietroPasotti, line 118: I'd add here a `_check_juju_supports_secrets` method call that verifies that we have the required hook tools

PietroPasotti: and raise a big exception if not

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -0,0 +1,391 @@
...
+                ):
+                    logger.error("Relation %s not found", self.certificates_relation_name)
+                    return
+
+                secret = self.charm.unit.add_secret(content, label="ca-certificate-chain")
+                secret.grant(relation)
+                relation.data[self.charm.unit]["secret-id"] = secret.id  # pyright: ignore
+                self.on.cert_changed.emit()  # pyright: ignore
+
+    def _retrieve_from_secret(self, value: str, secret_id_name: str):
```
PietroPasotti, line 287: add return type annotation

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -0,0 +1,391 @@
...
+            return None
+
+        if not (secret := self.model.get_secret(id=secret_id)):
+            return None
+
+        content = secret.get_content()
+        return content.get(value)
+
+    @property
+    def key(self):
```
PietroPasotti, line 301: why not `private_key`?

PietroPasotti: I never like it when
```
def foo():
    return _bar()
````

then either `bar` is a misnomer, or `foo` is.

File: lib/charms/observability_libs/v1/cert_handler.py [hunk 9]
```diff
@@ -0,0 +1,391 @@
...
+
+        secret = self.model.get_secret(id=secret_id)
+        secret.set_content({"csr": value})
+
+    @property
+    def _ca_cert(self) -> Optional[str]:
+        return self._retrieve_from_secret("ca-cert", "secret-id")
+
+    @property
+    def cert(self):
```
PietroPasotti, line 332: same as above, perhaps we can take this chance to uniform the naming of the methods (I presume you didn't change them)

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -0,0 +1,391 @@
...
+            secret.grant(relation)
+            relation.data[self.charm.unit]["csr-secret-id"] = secret.id  # pyright: ignore
+            return
+
+        secret = self.model.get_secret(id=secret_id)
+        secret.set_content({"csr": value})
+
+    @property
+    def _ca_cert(self) -> Optional[str]:
+        return self._retrieve_from_secret("ca-cert", "secret-id")
```
PietroPasotti, line 329: I think `secret-id` should be a global, instead of hardcoding it here and there
Something like:

`SECRET_ID_RELATION_KEY_NAME = "secret-id"`

and then use it directly in `_retrieve_from_secret`. I think the chances that someone will override or change this key are slim.

PietroPasotti (changes requested):
A few points that should be addressed, but it looks like it's going in the right direction.

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -0,0 +1,391 @@
...
+        # We need to check for units as a temporary workaround because of https://bugs.launchpad.net/juju/+bug/2024583
+        # This could in theory not work correctly on scale down to 0 but it is necessary for the moment.
+        return (
+            len(self.charm.model.relations[self.certificates_relation_name]) > 0
+            and len(self.charm.model.get_relation(self.certificates_relation_name).units) > 0  # type: ignore
+        )
+
+    def _on_certificates_relation_joined(self, _) -> None:
+        """Generate the CSR and request the certificate creation."""
+        self._generate_privkey()
```
sed-i, line 173: In a separate issue, should we decouple privkey management from cert handling?

lib/charms/observability_libs/v1/cert_handler.py
Abuelodelanada: Yes, I did not change the names to keep as equal as possible the API... But now I think there is no point in keep them.

lib/charms/observability_libs/v1/cert_handler.py
Abuelodelanada: Let's follow this here: https://github.com/canonical/observability-libs/issues/67

lib/charms/observability_libs/v1/cert_handler.py
Abuelodelanada: I can create the constant for that, but I cannot use it directly in `_retrieve_from_secret`, since for instance we have:

`self._retrieve_from_secret("csr", "csr-secret-id")` and `self._retrieve_from_secret("csr", "csr-secret-id")`

lucabello (approved):

PR #73, opened 2024-02-12
raise error on duplicate ports passed to patch

PietroPasotti:
## Issue
the service patch silently fails if you pass duplicate ports to it, say
```
                ports=[
                    ServicePort(name="svc1", port=1234, protocol="TCP"),
                    ServicePort(name="svc2", port=1234, protocol="TCP"),
                ],
```

## Solution

catch that and raise an exception as that's user error

## Context
this would have saved me a couple of hours of debugging this week


simskij:
Since we're moving to using the ops facility for this, the change is no longer needed. We should instead look to deprecate this library.

lucabello:
Closing per the comment above.

Issue #75, opened 2024-02-13
on config changed we are not creating the private key before generating the csr


Abuelodelanada:
### Enhancement Proposal

```python
unit-grafana-0: 17:24:10.771 ERROR unit.grafana/0.juju-log certificates:31: Uncaught exception while in charm code:
Traceback (most recent call last):
  File "./src/charm.py", line 1480, in <module>
    main(GrafanaCharm, use_juju_for_storage=True)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/main.py", line 456, in main
    _emit_charm_event(charm, dispatcher.event_name)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/main.py", line 144, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/framework.py", line 351, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/framework.py", line 853, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/framework.py", line 943, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-grafana-0/charm/lib/charms/tls_certificates_interface/v2/tls_certificates.py", line 1708, in _on_relation_broken
    self.on.all_certificates_invalidated.emit()
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/framework.py", line 351, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/framework.py", line 853, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/framework.py", line 943, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-grafana-0/charm/lib/charms/tempo_k8s/v1/charm_tracing.py", line 538, in wrapped_function
    return callable(*args, **kwargs)  # type: ignore
  File "/var/lib/juju/agents/unit-grafana-0/charm/lib/charms/observability_libs/v1/cert_handler.py", line 399, in _on_all_certificates_invalidated
    self._generate_csr(overwrite=True, clear_cert=True)
[...]

PR #76, opened 2024-02-19
fix race condition in privkey generation

lucabello:
## Issue

When trying to use this library in Alertmanager, I got the following error:

```
  File "/var/lib/juju/agents/unit-am-0/charm/lib/charms/observability_libs/v1/cert_handler.py", line 214, in _on_config_changed
    self._generate_csr(renew=True)
  File "/var/lib/juju/agents/unit-am-0/charm/lib/charms/observability_libs/v1/cert_handler.py", line 233, in _generate_csr
    raise RuntimeError(
RuntimeError: private key unset. call _generate_privkey() before you call this method.
```

In fact, `_on_config_changed()` doesn't call `_generate_privkey()`.

The issue is that a race condition probably causes **config-changed** to arrive before **certificates-relation-joined**, which doesn't generate the private key beforehand.

Please note that calling the function multiple times isn't a problem as the function will just skip :)

## Solution

Add a call to `_generate_privkey()` in the `_on_config_changed()` handler.

### DIFF
--- a/lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
+++ b/lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
@@ -133,7 +133,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 6
+LIBPATCH = 7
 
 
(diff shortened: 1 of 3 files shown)


PietroPasotti (approved):
LGTM

Abuelodelanada:
This PR solves: https://github.com/canonical/observability-libs/issues/75

Abuelodelanada (approved):

PR #82, opened 2024-03-26
Use tls-certificates v3

sed-i:
## Issue
Certificates generated by the charm don't include intermediates: https://github.com/canonical/httprequest-lego-k8s-operator/issues/93.

## Solution
This was addressed by https://github.com/canonical/tls-certificates-interface/pull/141 in v3, but wasn't back-ported to v2, so bump to v3.

Technically the diff presents a breaking change for the `.chain` property, but we haven't really used it anywhere yet.

In tandem with:
- https://github.com/canonical/traefik-k8s-operator/pull/317

## Upgrade notes
Due to this diff:
```diff
-if chain := self._peer_relation.data[self.charm.unit].get("chain", []):
+if chain := self._peer_relation.data[self.charm.unit].get("chain", ""):
```
on charm upgrade, the new charm would be expecting a `str`, but peer data would hold a `List[str]`. For this reason, first unrelate (tls_certificates), upgrade, and then relate again.

### DIFF
--- a/lib/charms/observability_libs/v0/cert_handler.py
+++ b/lib/charms/observability_libs/v0/cert_handler.py
@@ -45,7 +45,7 @@ try:
         CertificateAvailableEvent,
         CertificateExpiringEvent,
         CertificateInvalidatedEvent,
-        TLSCertificatesRequiresV2,
+        TLSCertificatesRequiresV3,
         generate_csr,
         generate_private_key,
     )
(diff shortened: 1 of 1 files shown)


Abuelodelanada (commented):
Shouldn't we use `TLSCertificatesRequiresV3` in cert_handler v1 too?

sed-i:
> Shouldn't we use `TLSCertificatesRequiresV3` in cert_handler v1 too?

I didn't bump cert_handler, it's still at v0.

Abuelodelanada (approved):

PR #83, opened 2024-03-27
Fix cert chain handling

sed-i:
## Issue
This PR:
- adds backwards compatibility to v0 in the change introduced in #82.
- apply the same change as #82 to v1.

## Solution
If "chain" is a list, convert it to str.


lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -278,7 +277,7 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
             content = {
                 "ca-cert": event.ca,
                 "server-cert": event.certificate,
-                "chain": json.dumps(event.chain),
+                "chain": event.chain_as_pem(),
```
sed-i, line 280: No need for backwards compat, because there was an omitted `json.loads`. I.e. the `chain` field is unused, otherwise we'd hear about it.

Abuelodelanada (approved):

PR #87, opened 2024-05-01
Certhandler secrets vault

PietroPasotti:
## Issue
certhandler is overly complicated, storing public data in too many juju secrets

## Solution
- remove two of the three secrets, only store the privkey in a secret and retrieve csr and chain as needed from the relation data
- instead of storing secret ids in relation databags, rely on hardcoded secret labels to always know what secret to pick
- generate privkey on-demand and use the secret storage to persist and verify if we've generated it already
- automatically detect if juju supports secrets, and if not fall back to a 'peers' peer-relation backed storage solution for the private key (as plaintext).
- automatically migrate from peer to secrets- backend if on charm-upgrade we obtain access to secrets

## Testing Instructions
A good test would be:
- deploy cos-lite
- jhack sync the lib to all charms who use it
- deploy self-signed certs and imatrix fill

Tandem pr: https://github.com/canonical/traefik-k8s-operator/pull/346


sed-i:
Great @PietroPasotti.
Can you link to the tandem PR?

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -87,15 +89,164 @@ class CertHandlerEvents(ObjectEvents):
     cert_changed = EventSource(CertChanged)
 
 
+class _VaultBackend(abc.ABC):
```
sed-i, line 92: Should we probably outline the "spec" in a docstring? For example:
```suggestion
class _VaultBackend(abc.ABC):
    """Base class for a single secret manager.
    
    Assumptions:
    - A single secret (label) is managed by a single instance.
    - Secret is per-unit (not per-app, i.e. may differ from unit to unit).
    """
```

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -87,15 +89,164 @@ class CertHandlerEvents(ObjectEvents):
...
+    def store(self, contents: Dict[str, str], clear: bool = False): ...
+
+    def get_value(self, key: str) -> Optional[str]: ...
+
+    def retrieve(self) -> Dict[str, str]: ...
```
sed-i, line 97: If we manage just one secret (label) then this method is redundant.

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -87,15 +89,164 @@ class CertHandlerEvents(ObjectEvents):
...
+    """Relation backend for Vault.
+
+    Use it to store data in a relation databag.
+    Assumes that a single relation exists and its data is readable.
+    If not, it will raise RuntimeErrors as soon as you try to read/write.
+    It will store the data, in plaintext (json-dumped) nested under a configurable
+    key in the **unit databag** of this relation.
+    """
+
+    def __init__(self, charm: CharmBase, relation_name: str, nest_under: str = "secret-contents"):
```
sed-i, line 112: "Vault" is private to cert_handler so I imagine we could avoid exposing `nest_under` as an arg. Upgrades would be easier if we hardcode the key. For example we could set it to 
```python
self.nest_under = "lib.charms.observability_libs.v1.cert_handler::secret-contents"
```
It would still be an unenforceable convention, but everything is.

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -87,15 +89,164 @@ class CertHandlerEvents(ObjectEvents):
...
+        self._databag[self.nest_under] = json.dumps(value)
+
+    def store(self, contents: Dict[str, str], clear: bool = False):
+        """Create a new revision by updating the previous one with ``contents``."""
+        current = self._read()
+
+        if clear:
+            current.clear()
+
+        current.update(contents)
```
sed-i, line 158: Do we need to `json.dumps` at any point? In `_read` we have `json.loads`.

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -87,15 +89,164 @@ class CertHandlerEvents(ObjectEvents):
...
+
+    def get_value(self, key: str):
+        """Like retrieve, but single-value."""
+        return self._read().get(key)
+
+    def retrieve(self):
+        """Return the full vault content."""
+        return self._read()
+
+    def nuke(self):
```
sed-i, line 169: Surprisingly, `woke` is offended by `dummy` or `sanity` more than by `nuke`.
```suggestion
    def clear(self):
```

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -87,15 +89,164 @@ class CertHandlerEvents(ObjectEvents):
...
+
+class _RelationVaultBackend(_VaultBackend):
+    """Relation backend for Vault.
+
+    Use it to store data in a relation databag.
+    Assumes that a single relation exists and its data is readable.
+    If not, it will raise RuntimeErrors as soon as you try to read/write.
+    It will store the data, in plaintext (json-dumped) nested under a configurable
+    key in the **unit databag** of this relation.
+    """
```
sed-i, line 110:
```suggestion
    
    Note: it is assumed that this object has exclusive access to the data, even though in practice it does not.
      Modifying relation data yourself would go unnoticed and disrupt consistency.
    """
```

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -87,15 +89,164 @@ class CertHandlerEvents(ObjectEvents):
...
+        del self._databag[self.nest_under]
+
+
+class _SecretVaultBackend(_VaultBackend):
+    """Relation backend for Vault.
+
+    Use it to store data in a Juju secret.
+    Assumes that Juju supports secrets.
+    If not, it will raise some exception as soon as you try to read/write.
+    """
```
sed-i, line 179:
```suggestion
    
    Note: it is assumed that this object has exclusive access to the data, even though in practice it does not.
      Modifying secret's data yourself would go unnoticed and disrupt consistency.
    """
```

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -130,6 +280,17 @@ def __init__(
         self.sans_ip = list(filter(is_ip_address, sans))
         self.sans_dns = list(filterfalse(is_ip_address, sans))
 
+        if self._check_juju_supports_secrets():
+            vault_backend = _SecretVaultBackend(charm, label="cert-handler-private-vault")
```
sed-i, line 284: The `label` arg should probably be in `Vault`, not in the backend.

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -161,6 +322,34 @@ def __init__(
...
+
+    def _on_upgrade_charm(self, _):
+        self._migrate_vault()
+
+    def _migrate_vault(self):
+        peer_backend = _RelationVaultBackend(self.charm, relation_name="peers")
+
+        if self._check_juju_supports_secrets():
+            # we are on recent juju
+            if self.vault.retrieve():
```
sed-i, line 338: `retrieve` is a tuple and always evaluates to true.
But we probably don't need `retrieve` anyway and should be using `get_value`?

lib/charms/observability_libs/v1/cert_handler.py
PietroPasotti: the difference between get value and retrieve is not the secret they retrieve from, but whether to retrieve the whole secret contents or just part of it. 

`get_value(value)` is shorthand for `retrieve().get(value)`

PietroPasotti: Vault wraps a single secret

lib/charms/observability_libs/v1/cert_handler.py
PietroPasotti: I meant it for generality, if someone ever wants to use multiple Vaults and/or _RelationVaultBackend instances and have them backed by the same relation, they'll need to use a separate key or they'll overwrite each other

PietroPasotti: but for now we can skip that and make a note, agreed

lib/charms/observability_libs/v1/cert_handler.py
PietroPasotti: probably fixed that already, there's now a corresponding dumps in _write.

lib/charms/observability_libs/v1/cert_handler.py
PietroPasotti: Mmm I tend to disagree because it's secret-specific and Vault is backend-agnostic.
Renamed it to `secret_label` for clarity. 
Symmetrically, _RelationVaultBackend has a `relation_name` arg that is none of Vault's concern.

lib/charms/observability_libs/v1/cert_handler.py
PietroPasotti: `retrieve()` is a `Dict[str,str]` and will evaluate to False if it's empty.

Abuelodelanada (approved):

PietroPasotti:
@sed-i tandem pr: https://github.com/canonical/traefik-k8s-operator/pull/346

PietroPasotti:
Traefik's TLS itests passed. Merging away.

sed-i:
> Traefik's TLS itests [passed AKA failed but for the wrong reasons](https://github.com/canonical/loki-k8s-operator/actions/runs/8923195780/job/24506948109?pr=392). Merging away.

The linked itests are loki, not traefik. Traefik's itest failed for something that looks relevant:
https://github.com/canonical/traefik-k8s-operator/actions/runs/8925027611/job/24512805931?pr=346#step:6:1131

PR #90, opened 2024-05-10
KubernetesServicePatch Add support for creating a new K8s service

IbraAoad:
## Issue
Should be part of fixing https://github.com/canonical/traefik-k8s-operator/issues/319

## Solution
Adjust the lib logic to create a new k8s service instead of patching the juju default one if `service_type` is `loadbalancer`

### Tandem PR
https://github.com/canonical/traefik-k8s-operator/pull/356

### DIFF
--- a/lib/charms/observability_libs/v1/kubernetes_service_patch.py
+++ b/lib/charms/observability_libs/v1/kubernetes_service_patch.py
@@ -205,11 +205,13 @@ class KubernetesServicePatch(Object):
         """
         super().__init__(charm, "kubernetes-service-patch")
         self.charm = charm
-        self.service_name = service_name if service_name else self._app
+        self.service_name = service_name or self._app
+        if self.service_name == self._app and service_type == "LoadBalancer":
+            self.service_name= f'{self._app}-lb'
         self.service_type = service_type
         self.service = self._service_object(
             ports,
-            service_name,
+            self.service_name,
             service_type,
             additional_labels,
             additional_selectors,
@@ -222,7 +224,7 @@ class KubernetesServicePatch(Object):
         self.framework.observe(charm.on.install, self._patch)
         self.framework.observe(charm.on.upgrade_charm, self._patch)
         self.framework.observe(charm.on.update_status, self._patch)
-        self.framework.observe(charm.on.remove, self._remove_service)
+        self.framework.observe(charm.on.stop, self._remove_service)
 
         # apply user defined events
         if refresh_event:
@@ -258,8 +260,6 @@ class KubernetesServicePatch(Object):
         Returns:
             Service: A valid representation of a Kubernetes Service with the correct ports.
         """
-        if not service_name:
-            service_name = self._app
         labels = {"app.kubernetes.io/name": self._app}
         if additional_labels:
             labels.update(additional_labels)
@@ -301,12 +301,7 @@ class KubernetesServicePatch(Object):
                 if not self.service_type == "LoadBalancer":
                     self._delete_and_create_service(client)
                 else:
-                    self._delete_and_create_lb_service(client)
-            elif self.service_type == "LoadBalancer":
-                logger.error(
-                    "Failed to create service: cannot create a lb with the same name as the default service."
-                )
-                return
+                    self._create_lb_service(client)
             client.patch(Service, self.service_name, self.service, patch_type=PatchType.MERGE)
         except ApiError as e:
             if e.status.code == 403:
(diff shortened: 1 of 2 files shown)


lib/charms/observability_libs/v1/kubernetes_service_patch.py
```diff
@@ -160,6 +179,7 @@ def __init__(
         ports: List[ServicePort],
         service_name: Optional[str] = None,
         service_type: ServiceType = "ClusterIP",
+        is_new_service: Optional[bool] = False,
```
ca-scribner, line 182: (ignore me if this has been discussed elsewhere) this feels like adding a new concern to `KubernetesServicePatch`.  That KubernetesService**Patch** creates a new service feels odd, especially since the docstring says """A utility for patching the Kubernetes service set up by Juju."  

What about using something standalone for creating new services?

lib/charms/observability_libs/v1/kubernetes_service_patch.py
```diff
@@ -294,6 +320,16 @@ def _delete_and_create_service(self, client: Client):
         client.delete(Service, self._app, namespace=self._namespace)
         client.create(service)
 
+    def _delete_and_create_new_service(self, client: Client):
+        try:
+            service = client.get(Service, self.service_name, namespace=self._namespace)
```
ca-scribner, line 325: Should there be a guard here against using app_name as the service name?  I see the note in the docstring above about avoiding app_name since the default juju service uses that name, but afaict nothing actually enforces that suggestion.  That sounds like an easy way for the user to make a mistake

lib/charms/observability_libs/v1/kubernetes_service_patch.py
sed-i: Currently, we have `ServiceType = Literal["ClusterIP", "LoadBalancer"]`, but we no longer use this lib for the "ClusterIP" use case, because we have ops open_port.

That being said, at first glance it seems to me that we can remove the `is_new_service` arg entirely, and always create a new service for LoadBalancer (but keep patching if ClusterIP).
This way we maintain backwards compat, while solving the LoadBalancer issue with no additional code changes beyond fetch-lib.
Wdyt?

IbraAoad: Thanks this definitely looks a lot better, it should be updated now

ca-scribner: To me it still feels odd that something labeled "patch" creates new things - I don't think someone coming to this lib would expect the behaviour.  But I also respect this is a low-code way of addressing the issue for traefik.

lib/charms/observability_libs/v1/kubernetes_service_patch.py
```diff
@@ -294,6 +323,16 @@ def _delete_and_create_service(self, client: Client):
         client.delete(Service, self._app, namespace=self._namespace)
         client.create(service)
 
+    def _delete_and_create_lb_service(self, client: Client):
+        try:
+            service = client.get(Service, self.service_name, namespace=self._namespace)
+            service.metadata.name = self.service_name  # type: ignore[attr-defined]
+            service.metadata.resourceVersion = service.metadata.uid = None  # type: ignore[attr-defined]   # noqa: E501
+            client.delete(Service, self.service_name, namespace=self._namespace)
+            client.create(service)
```
ca-scribner, line 332: Is there a reason to delete this at all?  iiuc `_delete_and_create_service` does the delete+create to allow for renaming (it `client.get()`s the service called `self._app`, then creates a service called `self.service_name`).  Here, we get, "change" the name (but replacing it with the same name we `client.get()`), then create.  This is effectively a noop

IbraAoad: you're right, I may have duplicated the logic from `_delete_and_create_service` assuming my goal is the same which apparently isn't, it should be updated now not to delete the lb and only create it if it's not there

regarding your other comment, even tho it's not instant in all my tests (on local and on CI) the ip was always generated way before the pebble ready event, I think in cases of a public provider it might take a while to provision and will indeed introduce a period of unavailability @sed-i @simskij wdyt?

simskij: Yes, it should not delete the lb if it exists as that would defeat the point for making the change in the first place. 

> I think too that externalip generation for a loadbalancer is not instant, but can instead take some time for something to be provisioned. So this might also be introducing periods where the service is unavailable

The juju-created service will still be there, but yes - you're right. The service will be unavailable for as long as it takes for traefik to get an external IP. That is not avoidable.

lucabello (approved):
All comments have been addressed and it looks good to me; a manual test with traefik would probably be great before merging :)

PR #91, opened 2024-05-14
add check for private key existence

michaeldmitry:
## Issue
by simply getting the `private_key` property, it now generates a new private key if it doesn't exist. Some charms access that property as a way of checking if its been generated or not (alertmanager), so it falsely generates a private key for the charm without even enabling tls on that charm.

## Solution
added a new function to check for the existence of the key

### DIFF
--- a/lib/charms/observability_libs/v1/cert_handler.py
+++ b/lib/charms/observability_libs/v1/cert_handler.py
@@ -477,6 +477,14 @@ class CertHandler(Object):
         """Emit cert-changed."""
         self.on.cert_changed.emit()  # pyright: ignore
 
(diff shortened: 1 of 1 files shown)


PietroPasotti (commented):
If we add this property, we still have to modify all charms that use `if cert.private_key` to say `if cert.private_key_exists` instead.
IMHO we should instead modify them so that they don't rely on the private key to verify that they have TLS at all. They should be checking if they have a cert, that's all. Privkey is a requirement for having a cert.
The check could be:
`if certhandler.cert`

this way, it will only return true if it has a cert. And if it has a cert, it's safe to assume it also has a privkey.

PR #99, opened 2024-06-17
[cert handler] do not observe rel broken directly

sed-i:
## Issue
We attempt to generate a CSR on relation broken.

## Solution
- Do not observe rel broken directly
- Do not generate CSR on rel broken

Addresses https://github.com/canonical/istio-operators/issues/441.


PietroPasotti (changes requested):
I'd really like to see at least one test to guard against regressions

PR #107, opened 2024-08-20
Transform `KubernetesComputeResourcesPatch` "Push" statuses into "Pull" statuses

michaeldmitry:
## Issue
The way how v0 `KubernetesComputeResourcesPatch` currently  behaves enforces charms using the library to implement a `push` status pattern to keep track of failures that might have happened during the patching. Even the existing function `is_ready` doesn't have enough info to determine whether the patch is not ready because it has failed or because its still in progress. That introduces an overhead of managing that "push" status and probably maintaining a `StoredState` in the charm. 

## Solution
Introduce a new API function `get_status` that gets the current status of the patch operation.
`get_status` will return one of 3 possible states:
1- succeeded: a patch operation has completed successfully or the patch operation itself has not started yet. 
2- failed: using `dry-run` mechanism, detect if the patch operation would've failed.
3- in_progress: `rollout status`-like logic to detect if the patch is still being applied and not fully ready.

## Testing Instructions
A tandem PR for [loki](https://github.com/canonical/loki-k8s-operator/pull/443) for using the new version of the library and test existing functionality as normal.

## Upgrade Notes
Charms can continue using this new version of the library same way as before relying on `is_ready` as a push status. If charms wish to transform that as a pull status, they should start using `get_status` in `collect-unit-status` and depending on the state of the patch returned, set the desired state of the charm.


lib/charms/observability_libs/v1/kubernetes_compute_resources_patch.py
```diff
@@ -0,0 +1,631 @@
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""# KubernetesComputeResourcesPatch Library.
+
+This library is designed to enable developers to more simply patch the Kubernetes compute resource
+limits and requests created by Juju during the deployment of a sidecar charm.
```
PietroPasotti, line 7: the reference to 'sidecar charm' is kind of old-fashioned. It's just a charm these days.

lib/charms/observability_libs/v1/kubernetes_compute_resources_patch.py
```diff
@@ -0,0 +1,631 @@
...
+        )
+        return podspec.resources
+
+    def is_ready(self, pod_name, resource_reqs: ResourceRequirements):
+        """Reports if the resource patch has been applied and is in effect.
+
+        Returns:
+            bool: A boolean indicating if the service patch has been applied and is in effect.
+        """
+        logger.info(
```
PietroPasotti, line 413: debug is perhaps enough? Not sure how large this message can get but I suppose "quite"

PietroPasotti: also perhaps we should only be logging on False return value? Not sure what's the point of logging all the data if things are well

lib/charms/observability_libs/v1/kubernetes_compute_resources_patch.py
```diff
@@ -0,0 +1,631 @@
...
+        return self.is_patched(resource_reqs) and equals_canonically(  # pyright: ignore
+            resource_reqs, self.get_actual(pod_name)  # pyright: ignore
+        )
+
+    def apply(self, resource_reqs: ResourceRequirements) -> None:
+        """Patch the Kubernetes resources created by Juju to limit cpu or mem."""
+        # Need to ignore invalid input, otherwise the StatefulSet gives "FailedCreate" and the
+        # charm would be stuck in unknown/lost.
+        if self.is_patched(resource_reqs):
+            return
```
PietroPasotti, line 428: I'd give the caller some indication of whether we did something or not. 
Either add a bool return value (True -> patch applied, we did do something, False -> no-op)
Or we could raise a custom exception so the user knows they should be checking if the patch is applied already.

lib/charms/observability_libs/v1/kubernetes_compute_resources_patch.py
```diff
@@ -0,0 +1,631 @@
...
+                    self._charm.unit.status = MaintenanceStatus(
+                        f"retrying patching resource limit... (attempt #{attempt.retry_state.attempt_number})"
+                    )
+                    self.patcher.apply(resource_reqs)
+
+        except exceptions.ConfigError as e:
+            msg = f"Error creating k8s client: {e}"
+            logger.error(msg)
+            self.on.patch_failed.emit(message=msg)
+            raise exceptions.ConfigError(msg) from e
```
PietroPasotti, line 541: a bit confused here by why we're reraising the same exception type but with a different message. If the user sees `lightkube.exceptions.ConfigError`, they'd  rightfully assume that the exception originates from lightkube while in fact you're firing it. 
I'd be in favour of doing a `raise PatchFailedError(msg) from e` instead.

PietroPasotti: or simply reraise it without changing the message. 
Also FYI you can:
`logger.exception(msg)` which will automatically attach the traceback to the error log.

PietroPasotti (commented):
A few thoughts and doubts about the public API. 
Didn't do an in-depth review and didn't try it out yet. Could you open a tandem PR in some charm where you replace the old lib with the new one, so we can clone that, deploy it, and see it in action?

Wondering if it makes sense to only move the files before merging but not while in the review stage? Having this large diff makes it really hard to look at the code.

lib/charms/observability_libs/v1/kubernetes_compute_resources_patch.py
michaeldmitry: the caller will get notified by an exception if things go sideways.
the caller interprets a "no exceptions fired" as an "everything is good".
its an internal call that will be handled by the caller function to nicely inform the user about it.

michaeldmitry: Added a debug log if its already patched

lib/charms/observability_libs/v1/kubernetes_compute_resources_patch.py
michaeldmitry: raised a custom `PatchFailedError` exception

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -468,51 +489,67 @@ def _on_config_changed(self, _):
...
             resource_reqs = self.resource_reqs_func()
             limits = resource_reqs.limits
             requests = resource_reqs.requests
         except ValueError as e:
             msg = f"Failed obtaining resource limit spec: {e}"
-            logger.error(msg)
-            self.on.patch_failed.emit(message=msg)
-            return
+            logger.exception(msg)
+            raise ValueError(msg) from e
```
sed-i, line 507: `_patch` is [called on config-changed](https://github.com/canonical/observability-libs/blob/a6ac58eb30286d34b269a1c585e5c8cc894ae373/lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py#L488), which means there is no obvious place to place a try-except guard in the charm: a charm instantiates the resource patch obj, which calls and observe, which goes into a queue, and only later, when ops processes the events, we'll get an exception.

I.e. it seems that with this design charm authors would need to wrap the call to `ops.main` with try-except, or intentionally/knowingly never try to catch it. I.e. semantically this is similar to [SystemExit](https://docs.python.org/3/library/exceptions.html#SystemExit). Are we sure we want to introduce this kind of behavior?

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -468,51 +489,67 @@ def _on_config_changed(self, _):
...
 
         for spec in (limits, requests):
             if not is_valid_spec(spec):
                 msg = f"Invalid resource limit spec: {spec}"
                 logger.error(msg)
-                self.on.patch_failed.emit(message=msg)
-                return
+                raise ValueError(msg)
```
sed-i, line 513: Should we have `PatchFailedError` everywhere instead of repeating ValueError?

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -468,51 +489,67 @@ def _on_config_changed(self, _):
...
-            self.on.patch_failed.emit(message=msg)
+            logger.exception(msg)
+            raise PatchFailedError(msg) from e
 
         except ValueError as e:
             msg = f"Kubernetes resources patch failed: {e}"
-            logger.error(msg)
-            self.on.patch_failed.emit(message=msg)
+            logger.exception(msg)
+            raise PatchFailedError(msg) from e
```
sed-i, line 552: Iiuc, raising the exception here, or in this PR in general, does not convert the status from push to pull.
Seems like the PR added retry which could help, but I don't see how this obviates the need for a stored state.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
michaeldmitry: The intention is not to catch it. Its to let it go to error state, as if you ever reach that point, there's no point of catching the error. The charm can do nothing and you'll need to manually fix sth.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
michaeldmitry: now, you only need to query `is_ready` to get the status of your **successful** patch. The pushing thing was mainly due to failures, right? Now, you don't have to wait for the event of failure, you'll know right away.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -422,6 +440,9 @@ class KubernetesComputeResourcesPatch(Object):
     """A utility for patching the Kubernetes compute resources set up by Juju."""
 
     on = K8sResourcePatchEvents()  # pyright: ignore
+    PATCH_RETRY_STOP = tenacity.stop_after_delay(60 * 3)
+    PATCH_RETRY_WAIT = tenacity.wait_fixed(30)
```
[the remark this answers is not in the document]
sed-i: If the charm sleeps (does not return) for 30sec, then juju will conclude the agent is down and will elect another leader. This is not a good thing. You can experiment with e.g. a 31 sec sleep to see the exact cascade.
In fact you could write a discourse post about your findings, I'm sure many will find it interesting.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
sed-i: I'm afraid users would not like it:
- Event handlers are processed by the order they were observed. What if there are other config-changed concerns that are now not being executed because our `observe` call came first? Do we really want an error state just because a resource limit failed to apply? Not sure this justifies a total outage.
- We support refresh events so the above may happen for other events, event custom events, that are passed in.

I think a SystemExit-like side effect from a charm lib is a big deviation from what most expect from charm libs. For example, one convention that we imposed on ourselves is that charmlibs shall not write unit status.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
sed-i: You mean the charm will crash until the resource patch works?
In that case we don't need `is_ready` at all :D

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -435,13 +533,36 @@ def apply(self, resource_reqs: ResourceRequirements) -> None:
...
+
+        patch_url = f"{self.api_url}/apis/apps/v1/namespaces/{self.namespace}/statefulsets/{self.statefulset_name}"
+
+        headers = {
+            "Authorization": f"Bearer {token}",
+            "Content-Type": "application/apply-patch+yaml",
+            "Accept": "application/json",
+        }
+
+        with httpx.Client(verify=self.ca_path) as client:
```
sed-i, line 550: Why do we need httpx?
I.e. we were able to use lighkube.Client before, can we continue using lightkube instead of httpx directly?
(Seems like lightkube is able to figure out api url and token from the env on its own.)

If lightkube doesn't support the dryRun param yet, would it be straightforward to contribute the feature there?

sed-i (commented):
Partial review for now, sorry :)

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
michaeldmitry: Refactored logic to not have any error states. Updated PR description.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
michaeldmitry: refactored and updated PR description

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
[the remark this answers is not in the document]
michaeldmitry: Will verify the above as a separate activity from this PR. for the purpose of this PR, set the retry value to `20` seconds to avoid any possible risks.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -435,13 +533,36 @@ def apply(self, resource_reqs: ResourceRequirements) -> None:
...
+
 
 class KubernetesComputeResourcesPatch(Object):
     """A utility for patching the Kubernetes compute resources set up by Juju."""
 
     on = K8sResourcePatchEvents()  # pyright: ignore
-    PATCH_RETRY_STOP = tenacity.stop_after_delay(60 * 3)
-    PATCH_RETRY_WAIT = tenacity.wait_fixed(30)
+    PATCH_RETRY_STOP = tenacity.stop_after_delay(20)
+    PATCH_RETRY_WAIT = tenacity.wait_fixed(5)
```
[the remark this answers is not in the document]
michaeldmitry: Will verify the above as a separate activity from this PR. for the purpose of this PR, set the retry value to 20 seconds to avoid any possible risks.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
michaeldmitry: lightkube's client doesn't have the support yet.
I opened an enhancement proposal https://github.com/gtsystem/lightkube/issues/73, which I can contribute to adding the support needed. 
But for the purpose of this PR, I think we can continue using httpx and add a TODO?

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -591,6 +712,36 @@ def is_ready(self) -> bool:
             self.on.patch_failed.emit(message=msg)
             return False
 
+    def get_status(self) -> Tuple[PatchState, str]:
```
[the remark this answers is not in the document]
michaeldmitry: yeah I thought about that approach, but I felt giving the charms the control over which state they want to set on the charm instead of kind of imposing it on them. But, on second thought, they still have that control if they want to change the state we see as suitable. 
Will change the return type.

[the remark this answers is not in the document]
tonyandrewmeyer: > This practice is not yet well established in the ecosystem, and there aren't any conventions yet really, and some may argue that `get_status` methods should return `Sequence[StatusBase]` in complex libs, but for now I think we can start with 1:1 match with `add_status`, implying that a lib should have only one status at a time, even if a few things went wrong.

I feel like this is a good approach. The lib should present its combined conclusion on its status, having logged (or aggregated into a message) anything extra needed. The charm is already going to need to combine the lib status with ones from other libs or its own requirements; it's better to not move more of the complexity into the charm and instead leave that responsibility with the lib.

I think this would also evolve well in the future, if there are per-relation statuses as has been mooted (with the assumption that a lib generally 'owns' one/more relations), where the lib would set a combined status for each relation - it seems unlikely it would be able to break it down further than that.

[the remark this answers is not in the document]
benhoyt: Yeah, I like @sed-i's suggestion here, makes it nice and easy for the caller to pass to `add_status()`, but doesn't actually call `add_status()`, so the caller could still disregard/tweak it if they wanted. I do think the simpler "single status" is better than "sequence of statuses".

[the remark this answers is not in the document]
dimaqq: I'd love to comment, but I think I need to understand the resource patch concept first.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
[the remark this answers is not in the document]
michaeldmitry: Short answer: nothing happens :)
Long answer:
https://discourse.charmhub.io/t/what-would-happen-if-a-charm-sleeps-for-a-long-time/15332

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
michaeldmitry: Merged in [v0.15.4](https://github.com/gtsystem/lightkube/releases/tag/v0.15.4)

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -417,11 +549,26 @@ def apply(self, resource_reqs: ResourceRequirements) -> None:
...
+        """Run a dry-run patch operation."""
+        self.client.patch(
+            StatefulSet,
+            self.statefulset_name,
+            self._patched_delta(resource_reqs),
+            namespace=self.namespace,
+            patch_type=PatchType.APPLY,
+            field_manager=self.__class__.__name__,
+            dry_run=True,
+        )
```
sed-i, line 562: Is there a way to consolidate this code with the real apply?
To avoid diverging in case we forget to update something in _both_ places.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -385,18 +415,119 @@ def get_actual(self, pod_name: str) -> Optional[ResourceRequirements]:
...
+
+        Implementation follows a similar approach to `kubectl rollout status statefulset` to track the progress of a rollout.
+        Reference: https://github.com/kubernetes/kubectl/blob/kubernetes-1.31.0/pkg/polymorphichelpers/rollout_status.go
+        """
+        try:
+            sts = self.client.get(
+                StatefulSet, name=self.statefulset_name, namespace=self.namespace
+            )
+        except (ValueError, ApiError):
+            return False
```
sed-i, line 470: Is (just) returning False here desired behavior?
Add a comment explaining?

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -385,18 +415,119 @@ def get_actual(self, pod_name: str) -> Optional[ResourceRequirements]:
...
+            return True, msg
+
+        return False, ""
+
+    def is_in_progress(self) -> bool:
+        """Returns a boolean to indicate whether a patch operation is in progress.
+
+        Implementation follows a similar approach to `kubectl rollout status statefulset` to track the progress of a rollout.
+        Reference: https://github.com/kubernetes/kubectl/blob/kubernetes-1.31.0/pkg/polymorphichelpers/rollout_status.go
+        """
```
sed-i, line 464: Should some of this functionality be added to lightkube?

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -406,6 +537,7 @@ def apply(self, resource_reqs: ResourceRequirements) -> None:
         # Need to ignore invalid input, otherwise the StatefulSet gives "FailedCreate" and the
         # charm would be stuck in unknown/lost.
         if self.is_patched(resource_reqs):
+            logger.debug(f"resource requests {resource_reqs} are already patched.")
```
sed-i, line 540: In favor of readability and grep-ability:
```suggestion
            logger.debug(f"Resource requests are already patched: {resource_reqs}")
```

sed-i (approved):
Great job!
The `is_in_progress` is difficult to grok/utest, and could probably be better to have in lightkube, but that's not a blocker for me.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
michaeldmitry: It definitely can!

PR #108, opened 2024-09-02
Add `refresh_events` to `CertHandler`

michaeldmitry:
This PR adds a `refresh_events` option to `CertHandler` which will be observed to check if there are any changes to the CSR (e.g: newly added SANs). And if there are changes from the existing CSR, it will be replaced by a new one generated with the newest request details and subsequently provide a new certificate.

## Context
In HA clusters, the coordinator currently requests a certificate with its own hostname given as a SAN. However, this certificate is passed on to the workers to be used by them. This leads to issues when services (e.g pebble check on the worker) tries to communicate with the worker server **directly**, they will get a certificate error saying that the certificate is valid for the coordinator not the worker. 
This solution gives the option to the coordinator to specify the workers' hostnames as SANS whenever a worker is added or removed by, for example, giving `refresh_events=[self.cluster.on.changed]` to check if the certificate needs to be refreshed.

### DIFF
--- a/lib/charms/observability_libs/v1/cert_handler.py
+++ b/lib/charms/observability_libs/v1/cert_handler.py
@@ -301,9 +301,9 @@ class CertHandler(Object):
             cert_subject: Custom subject. Name collisions are under the caller's responsibility.
             sans: DNS names. If none are given, use FQDN.
             refresh_events: an optional list of bound events which
-                will be observed to overwrite the current CSR with a newly generated one.
-                If there are no changes in the CSR request, no changes will happen to the
-                CSR or the certificate.
+                will be observed to replace the current CSR with a new one
+                if there are any changes in the CSR request. Then, subsequently, replace the
+                its corresponding certificate with a new one.
         """
         super().__init__(charm, key)
         self.charm = charm
--- a/lib/charms/observability_libs/v1/cert_handler.py
+++ b/lib/charms/observability_libs/v1/cert_handler.py
@@ -59,7 +59,7 @@ except ImportError as e:
 import logging
 
 from ops.charm import CharmBase
-from ops.framework import BoundEvent, EventBase, EventSource, Object, ObjectEvents
+from ops.framework import BoundEvent, EventBase, EventSource, Object, ObjectEvents, StoredState
 from ops.jujuversion import JujuVersion
 from ops.model import Relation, Secret, SecretNotFoundError
 
@@ -273,6 +273,7 @@ class CertHandler(Object):
     """A wrapper for the requirer side of the TLS Certificates charm library."""
 
     on = CertHandlerEvents()  # pyright: ignore
+    _stored = StoredState()
 
     def __init__(
         self,
@@ -302,10 +303,14 @@ class CertHandler(Object):
             sans: DNS names. If none are given, use FQDN.
             refresh_events: an optional list of bound events which
                 will be observed to replace the current CSR with a new one
-                if there are changes in the CSR's DNS SANs, IP SANs, subject, or private key.
+                if there are changes in the CSR's DNS SANs or IP SANs.
                 Then, subsequently, replace its corresponding certificate with a new one.
         """
         super().__init__(charm, key)
+        self._stored.set_default(
+            current_sans_ip=None,
+            current_sans_dns=None,
+        )
         self.charm = charm
 
         # We need to sanitize the unit name, otherwise route53 complains:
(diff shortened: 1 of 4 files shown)


lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -355,6 +360,22 @@ def __init__(
...
         )
 
+        if refresh_events:
+            for ev in refresh_events:
+                self.framework.observe(ev, self._on_refresh_event)
+
+    def _on_refresh_event(self, _):
+        # Renew only if there are CSR changes
+        curr_csr = self._csr.encode() if self._csr else None
+        new_csr = generate_csr(
```
PietroPasotti, line 370: it's not entirely clear to me why we need to generate a new csr in order to see if the one we have already is out of date. What things could possibly have changed? sans-ip, sans-dns, cert_subject, private-key, or all of the above?
I agree that it's easier to simply generate a new csr and compare with the old one instead of putting all these input params in, say, storedstate, and comparing them in order to determine if we need to renew our csr. But, we should be explaining in a comment or docstring why we're choosing this approach.

PietroPasotti (changes requested):
overall looks good, missing a little bit of comments/docstrings and ideally a test that demonstrates how a new csr is requested when you receive a refresh event and, say, your privkey has changed, and a new csr is NOT requested when all of the input params are unchanged.

lib/charms/observability_libs/v1/cert_handler.py
michaeldmitry: All fields' changes are valid to generate a new csr, even the private key. Added docstrings to explain better.

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -283,6 +283,7 @@ def __init__(
         peer_relation_name: str = "peers",
         cert_subject: Optional[str] = None,
         sans: Optional[List[str]] = None,
+        refresh_events: List[BoundEvent] = [],
```
Abuelodelanada, line 286: We shouldn't use mutable default argument.

```suggestion
        refresh_events: Optional[List[BoundEvent]] = None,
```

Abuelodelanada (changes requested):

lib/charms/observability_libs/v1/cert_handler.py
PietroPasotti: yup, I was suggesting using `()`

michaeldmitry: yeah, but `pyright` would complain. I'll keep it as `None` and add a check if `not None`

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -355,6 +360,27 @@ def __init__(
...
         )
 
+        if refresh_events:
+            for ev in refresh_events:
+                self.framework.observe(ev, self._on_refresh_event)
+
+    def _on_refresh_event(self, _):
+        """Replace the latest current CSR with a new one if there are any CSR changes.
+
+        The following CSR changes will trigger a certificate renewal: DNS SANs, IP SANs, subject, and private key changes.
```
PietroPasotti, line 370: ok, I think I see now, thanks for the clarification! do you think this behaviour would ever be good to have outside of the refresh event flow?
I mean, will the charm ever want to do something like: `self.cert_handler.refresh_csr()`?

PietroPasotti: if yes, we can pull this logic out to a separate (public?) method...

PietroPasotti (approved):

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -299,8 +301,16 @@ def __init__(
...
+            refresh_events: an optional list of bound events which
+                will be observed to replace the current CSR with a new one
+                if there are changes in the CSR's DNS SANs or IP SANs.
+                Then, subsequently, replace its corresponding certificate with a new one.
         """
         super().__init__(charm, key)
+        self._stored.set_default(
+            current_sans_ip=None,
+            current_sans_dns=None,
+        )
```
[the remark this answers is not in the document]
michaeldmitry: We need this to know what our current SANs that generated the current `_csr` to know if we need to generate a new CSR based on the SANs changes.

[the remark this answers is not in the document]
michaeldmitry: Peer data would assume the peer relation is defined in the charm using the library? Is there a concern about using storedState?

lib/charms/observability_libs/v1/cert_handler.py
michaeldmitry: > ok, I think I see now, thanks for the clarification! do you think this behaviour would ever be good to have outside of the refresh event flow? I mean, will the charm ever want to do something like: `self.cert_handler.refresh_csr()`?

Its possible, but do we need it? When certs approach expiry, for example, CertHandler would renew certs on its own without the charm's intervention. So, I think CertHandler is there to handle these sorts of interactions leaving out the need for a charm to manually invoke sth.

michaeldmitry: FYI
Comparing the entire CSR object will not work because the object would contain a UID that will always be different (Thanks to UTs for discovering this)
I refactored things to detect if renewal is needed based on SANs only while keeping the current SANs in a `StoredState`.

lib/charms/observability_libs/v1/cert_handler.py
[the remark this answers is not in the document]
PietroPasotti: in general the feeling is try to avoid it as much as possible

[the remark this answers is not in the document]
PietroPasotti: in traefik however we're already doing something similar, consider hashing all the bits you want to check for changes instead of storing them individually, it's more extensible https://github.com/canonical/traefik-k8s-operator/blob/main/src/charm.py#L130

PietroPasotti (approved):

lib/charms/observability_libs/v1/cert_handler.py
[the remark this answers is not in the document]
PietroPasotti: @sed-i this is blocking a bunch of other work so we'll go ahead with it. we can always open a ticket to refactor using stored data (or finding a way to compare csrs without storing their precursors) if you think it's worth it.

PR #109, opened 2024-09-13
Insert server certificate to chain if its not already there

michaeldmitry:
## Issue
The issue contributes to fix https://github.com/canonical/traefik-k8s-operator/issues/400.
Using a certificate provider, such as [manual tls certificates](https://github.com/canonical/manual-tls-certificates-operator) will not guarantee that the chain will contain the server certificate. And some services, such as, traefik[ depend on the chain to use as its server certificate](https://github.com/canonical/traefik-k8s-operator/blob/main/src/charm.py#L431).

## Solution
In property `chain`, add server certificate to the chain if its not already there.

## Context
https://github.com/canonical/traefik-k8s-operator/pull/317

## Testing Instructions
run `tox -e scenario-manual`


PietroPasotti (approved):
LGTM!

PR #110, opened 2024-10-10
Refresh CSR "whenever" needed

michaeldmitry:
## Issue
Currently, if SANs changed, we'd refresh the CSR only on the events passed to `refresh_events`. We should be doing that regardless of which event we're processing (as long as there was an initial CSR generated) since we might not know which specific events would trigger a SANs change.

## Solution
Add a "reconcile" logic in CertHandler constructor to refresh CSR if there has been any change to SANs.

## Testing Instructions
run `tox -e scenario-manual`

## Upgrade Notes
- Deprecation notice have been put if `refresh_events` is passed to CertHandler.

### DIFF
--- a/lib/charms/observability_libs/v1/cert_handler.py
+++ b/lib/charms/observability_libs/v1/cert_handler.py
@@ -318,6 +318,7 @@ class CertHandler(Object):
 
         # Use fqdn only if no SANs were given, and drop empty/duplicate SANs
         sans = list(set(filter(None, (sans or [socket.getfqdn()]))))
+        # sort SANS lists to avoid unnecessary csr renewals during reconciliation
         self.sans_ip = sorted(filter(is_ip_address, sans))
         self.sans_dns = sorted(filterfalse(is_ip_address, sans))
 
@@ -369,6 +370,10 @@ class CertHandler(Object):
                 "DEPRECATION WARNING. `refresh_events` is now deprecated. CertHandler will automatically refresh the CSR when necessary."
             )
 
+        self._reconcile()
+
+    def _reconcile(self):
+        """Run all logic that is independent of what event we're processing."""
         self._refresh_csr_if_needed()
 
     def _on_upgrade_charm(self, _):
@@ -385,7 +390,7 @@ class CertHandler(Object):
             self._generate_csr(renew=True)
 
     def _refresh_csr_if_needed(self):
-        """Refresh the latest current CSR with a new one if there are any SANs changes."""
+        """Refresh the current CSR with a new one if there are any SANs changes."""
         if self._stored.csr_hash is not None and self._stored.csr_hash != self._csr_hash:
             self._generate_csr(renew=True)
 
@@ -447,7 +452,11 @@ class CertHandler(Object):
         Only include here the config options that, should they change, should trigger a renewal of
         the CSR.
         """
-        return self._stable_hash(
+
+        def _stable_hash(data):
+            return hashlib.sha256(str(data).encode()).hexdigest()
+
+        return _stable_hash(
             (
                 tuple(self.sans_dns),
                 tuple(self.sans_ip),
@@ -639,6 +648,3 @@ class CertHandler(Object):
             logger.debug(msg)
             return False
         return True
-
-    def _stable_hash(self, data):
-        return hashlib.sha256(str(data).encode()).hexdigest()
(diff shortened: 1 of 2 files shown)


lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -320,8 +318,8 @@ def __init__(
 
         # Use fqdn only if no SANs were given, and drop empty/duplicate SANs
         sans = list(set(filter(None, (sans or [socket.getfqdn()]))))
-        self.sans_ip = list(filter(is_ip_address, sans))
-        self.sans_dns = list(filterfalse(is_ip_address, sans))
+        self.sans_ip = sorted(filter(is_ip_address, sans))
```
PietroPasotti, line 321: I'd add a comment to explain why we're sorting

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -367,13 +365,11 @@ def __init__(
...
-                self.framework.observe(ev, self._on_refresh_event)
+            logger.warn(
+                "DEPRECATION WARNING. `refresh_events` is now deprecated. CertHandler will automatically refresh the CSR when necessary."
+            )
 
-    def _on_refresh_event(self, _):
-        """Replace the latest current CSR with a new one if there are any SANs changes."""
-        if self._stored.csr_hash != self._csr_hash:
-            self._generate_csr(renew=True)
+        self._refresh_csr_if_needed()
```
PietroPasotti, line 372: since that seems to be the pattern, how about we call `self._reconcile` from init, and in `_reconcile` we call this method?
that way it'll be easier to extend when (not if) we want to add more stuff to the common exit hook

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -388,6 +384,11 @@ def _on_upgrade_charm(self, _):
             # this will call `self.private_key` which will generate a new privkey.
             self._generate_csr(renew=True)
 
+    def _refresh_csr_if_needed(self):
+        """Refresh the latest current CSR with a new one if there are any SANs changes."""
```
PietroPasotti, line 388:
```suggestion
        """Refresh the current CSR with a new one if there are any SANs changes."""
```

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -638,3 +639,6 @@ def _check_juju_supports_secrets(self) -> bool:
             logger.debug(msg)
             return False
         return True
+
+    def _stable_hash(self, data):
```
PietroPasotti, line 643: Imho this doesn't need to be a separate method. I vote we put this logic directly in `_csr_hash`, or if you feel like it should be its own function, either make it a staticmethod or make it a module-level helper function. No need to bind it to self.

tests/scenario/test_cert_handler/test_cert_handler_v1.py
```diff
@@ -183,48 +191,53 @@ def test_cert_joins_peer_vault_backend(ctx_juju2, certificates, leader):
         assert mgr.charm.ch.private_key
 
 
-def test_renew_csr_on_sans_change(ctx, certificates):
-    # generate a CSR
+@pytest.mark.parametrize(
+    "event,generate_call_count",
```
PietroPasotti, line 195:
```suggestion
    "event,expected_generate_calls",
```

tests/scenario/test_cert_handler/test_cert_handler_v1.py
```diff
@@ -183,48 +191,53 @@ def test_cert_joins_peer_vault_backend(ctx_juju2, certificates, leader):
...
+    (("update_status", 0), ("start", 0), ("install", 0), ("config_changed", 1)),
```
PietroPasotti, line 196: it's not entirely clear to me why we expect config_changed to result in a generate_cert call, but not the other events.

PietroPasotti (approved):
only a few nits from my side, doesn't need a second pass from me

tests/scenario/test_cert_handler/test_cert_handler_v1.py
michaeldmitry: because CertHandler generates the cert on the `config_changed` event https://github.com/canonical/observability-libs/blob/main/lib/charms/observability_libs/v1/cert_handler.py#L472

michaeldmitry: will add a comment with that on the test

PR #113, opened 2024-11-27
Validate LB annotations before reconciling 

IbraAoad:
## Issue
1 - Lib doesn't validate if annotations are changed to re-patch the service and only updates if ports are changed
2 - Lib currently uses `PatchType.MERGE` to update Kubernetes services, which doesn't handle removing old annotations when they are deleted.

## Solution
1 - Validate  `expected and fetched annotations` as part of `_is_patched` method.
2 - Use `PatchType.APPLY` to reconcile the annotations

## Context
Change is part of https://github.com/canonical/traefik-k8s-operator/issues/423 and is required to support adding custom annotations to load balancers.

## Testing Instructions
1 - Juju deploy traefik-k8s
2 - juju config traefik-k8s loadbalancer_annotations="key1=value1"
3 - kubectl describe traefik lb service and validate annotations

Tandem PR: https://github.com/canonical/traefik-k8s-operator/pull/428

### DIFF
--- a/lib/charms/observability_libs/v1/kubernetes_service_patch.py
+++ b/lib/charms/observability_libs/v1/kubernetes_service_patch.py
@@ -145,7 +145,7 @@ def setUp(self, *unused):
 
 import logging
 from types import MethodType
-from typing import Any, List, Literal, Optional, Union
+from typing import Any, Dict, List, Literal, Optional, Union
 
 from lightkube import ApiError, Client  # pyright: ignore
 from lightkube.core import exceptions
@@ -305,11 +305,10 @@ class KubernetesServicePatch(Object):
         try:
             if self._is_patched(client):
                 return
-            if self.service_name != self._app:
-                if not self.service_type == "LoadBalancer":
-                    self._delete_and_create_service(client)
-                else:
-                    self._create_lb_service(client)
+            if self.service_type == "LoadBalancer":
+                self._create_lb_service(client)
+            else:
+                self._delete_and_create_service(client)
             client.patch(
                 Service,
                 self.service_name,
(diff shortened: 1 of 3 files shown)


lib/charms/observability_libs/v1/kubernetes_service_patch.py
```diff
@@ -310,7 +310,13 @@ def _patch(self, _) -> None:
                     self._delete_and_create_service(client)
                 else:
                     self._create_lb_service(client)
-            client.patch(Service, self.service_name, self.service, patch_type=PatchType.MERGE)
+            client.patch(
+                Service,
+                self.service_name,
+                self.service,
+                patch_type=PatchType.APPLY,
```
ca-scribner, line 317: is there a reason why we'd want `APPLY` here but `STRATEGIC` (the default) in [istio-beacon](https://github.com/canonical/istio-beacon-k8s-operator/blob/c389ccc65d82d138197a222d20191b8e7e4b9317/src/charm.py#L362C35-L362C40)?

ca-scribner: I'm not sure this is an issue, but I'm raising it for discussion...

A consequence of using `APPLY` here is that lightkube will raise an exception if someone else already "owns" this object (ex: try doing two applies to an object using lightkube, each with a different field manager).  

This wouldn't be so bad if we owned the object, since we could be confident that nobody will set the field manager but us, but in this case Juju owns it.  So if Juju ever tries to apply to the object, we could have either of:
1. Juju sets field manager, so then our code will raise an exception here
2. our code sets field manager on apply, and if Juju ever tries to reconcile the object back to normal then Juju will hit an exception

Probably (1) isn't an issue (because we'd see it in integration tests), but (2) feels like an implementation detail that Juju owns.

ca-scribner (commented):
This LGTM apart from the one comment that might have wider implications

lib/charms/observability_libs/v1/kubernetes_service_patch.py
IbraAoad: Yeah, you're correct, I was under the impression that juju won't own this as it's an additional LB that we create, but I'm realizing now that you can also request a `ClusterIP` service with this lib which will override a juju owned object.

Not sure if `Strategic` patch will give us what we want but will try it out.

IbraAoad:
After discussions with @simskij and @dstathis , this PR will be closed in favor of #116. The Traefik charm will no longer use the library and will instead incorporate the necessary logic for creating its LoadBalancer directly into the charm code.

PR #116, opened 2024-12-03
Deprecate `kubernetes_service_patch` lib

michaeldmitry:
## Issue
This library has been used to patch the Kubernetes service created by Juju during the deployment of a charm to open ports on the unit and such. This functionality is available in Ops, as part of [ops.Unit.set_ports](https://ops.readthedocs.io/en/latest/#ops.Unit.set_ports). ~~However, this library is still used by charms like `traefik-k8s` to provide loadbalancing capabilities by creating/patching k8s LB services.~~

## Solution
- Add a deprecation warning and that ops.Unit.set_ports should be used instead.
~~- Update docstring to mention the same as well as that the library is now only intended to be used for creating/patching LoadBalancer k8s services.~~

### DIFF
--- a/lib/charms/observability_libs/v1/kubernetes_service_patch.py
+++ b/lib/charms/observability_libs/v1/kubernetes_service_patch.py
@@ -7,7 +7,7 @@
 The `kubernetes_service_patch` library is DEPRECATED for patching the Kubernetes service created
 by Juju during the deployment of a charm and `ops.Unit.set_ports` functionality should be used for that instead.
 
-This library is intended to only be used to enable developers to create/patch Kubernetes ``LoadBalancer`` services.
+This library is intended to only be used to enable developers to create/patch Kubernetes `LoadBalancer` services.
 
 When initialised, this library binds a handler to the parent charm's `install` and `upgrade_charm`
 events which applies the patch to the cluster. This should ensure that the service ports are
--- a/lib/charms/observability_libs/v1/kubernetes_service_patch.py
+++ b/lib/charms/observability_libs/v1/kubernetes_service_patch.py
@@ -3,14 +3,11 @@
 
 """# KubernetesServicePatch Library.
 
-This library is designed to enable developers to more simply patch the Kubernetes Service created
-by Juju during the deployment of a sidecar charm. When sidecar charms are deployed, Juju creates a
-service named after the application in the namespace (named after the Juju model). This service by
-default contains a "placeholder" port, which is 65535/TCP.
+[DEPRECATED!]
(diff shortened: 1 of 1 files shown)


simskij:
@michaeldmitry If the LB creation functionality is indeed only used in traefik, can't we just copy the code there and deprecate the whole lib?

michaeldmitry:
> @michaeldmitry If the LB creation functionality is indeed only used in traefik, can't we just copy the code there and deprecate the whole lib?

@simskij not only by traefik. its used by [mlflow-operator](https://github.com/canonical/mlflow-operator), [consul-k8s-operator](https://github.com/canonical/consul-k8s-operator/tree/a5ff32e42aeab37a66fb95bfcaeaa21b10aecaf5), [several sunbeam charms](https://opendev.org/openstack/sunbeam-charms/src/branch/main/charms) for similar purposes as traefik

lib/charms/observability_libs/v1/kubernetes_service_patch.py
```diff
@@ -3,14 +3,11 @@
 
 """# KubernetesServicePatch Library.
 
-This library is designed to enable developers to more simply patch the Kubernetes Service created
-by Juju during the deployment of a sidecar charm. When sidecar charms are deployed, Juju creates a
-service named after the application in the namespace (named after the Juju model). This service by
-default contains a "placeholder" port, which is 65535/TCP.
+[DEPRECATED!]
+The `kubernetes_service_patch` library is DEPRECATED for patching the Kubernetes service created
+by Juju during the deployment of a charm and `ops.Unit.set_ports` functionality should be used for that instead.
```
dstathis, line 8: After discussion with @simskij and @IbraAoad, We think that we should deprecate the entire library, not just the part pertaining to juju owned resources.

dstathis: Also, we should add to the deprecation a removal date of October 2025.

lib/charms/observability_libs/v1/kubernetes_service_patch.py
```diff
@@ -167,13 +150,13 @@ def setUp(self, *unused):
...
 # to 0 if you are raising the major API version
-LIBPATCH = 12
+LIBPATCH = 13
 
 ServiceType = Literal["ClusterIP", "LoadBalancer"]
 
 
 class KubernetesServicePatch(Object):
-    """A utility for patching the Kubernetes service set up by Juju."""
+    """A utility for patching a Kubernetes /NodePort service."""
```
dstathis, line 159: Probably we don't need these doc changes if we deprecate the entire library.

IbraAoad (approved):

simskij (approved):

PR #123, opened 2025-06-05
feat: deprecate juju_topology v0

lucabello:
## Issue
`juju_topology` v0 is an old library we don't touch anymore. The current `juju_topology` we use is in cos-lib: 
https://github.com/canonical/cos-lib/blob/6cd3769727419f50018994591a90c0024542389b/src/cosl/juju_topology.py

## Solution
Add a deprecation warning, as an intermediate step before removing the library entirely.


MichaelThamm (approved):
LGTM

PR #124, opened 2025-06-27
fix: chain ordering see https://github.com/canonical/traefik-k8s-oper…

dstathis:
## Issue
canonical/traefik-k8s-operator#491

## Solution
We need to use a new method that returns the chain in the reverse order.


PietroPasotti (approved):
Looks right, but I'd add a chonky comment to explain why we're using this method specifically, for when we forget about this.

MichaelThamm:
> Looks right, but I'd add a chonky comment to explain why we're using this method specifically, for when we forget about this.

This method is just the updated version of the original. Not really sure we need a chunky block comment since its doing the same thing, just in the right way.

PR #125, opened 2025-07-09
make the chain order fix backwards compatible

dstathis:
This improves upon #124 to make it backwards compatible with older revisions of self-signed-certificates. It does this by checking if the leaf is first, and if not, reversing the chain.


MichaelThamm:
Prior to this PR being merged, the working version combinations are:

Traefik rev237 & SSC >=rev309
Traefik <=rev236 & SSC <=rev308

What will be the compatible version ocmbinations after merging?

dstathis:
> Prior to this PR being merged, the working version combinations are:
> 
> Traefik rev237 & SSC >=rev309
> Traefik <=rev236 & SSC <=rev308
> 
> What will be the compatible version ocmbinations after merging?

This revision of Traefik will be compatible with all versions of SSC.

Issue #128, opened 2025-08-21
k8s resource patch is losing the fight with juju


PietroPasotti:
### Bug Description

In some charms, we apply resource constraints on the charm container because we wanted fine-grained control over how much memory/cpu the charm container got without having to apply the same to the workload container(s).

juju 3.6.9 started setting a default memory limit (64M) **on the charm container**. 
This breaks the service patch because we now get conflicts when we try to override it.
Our goal is to have per-container resource requests.

AFFECTED CHARMS:
🐛 loki-coordinator-k8s
🐛 mimir-coordinator-k8s
🐛 pyroscope-operators
🐛 tempo-operators

### To Reproduce

deploy anything using the service patch on juju 3.6.9 

### Environment

juju 3.6.9 

### Relevant log output

```shell
iunit-tempo-0: 10:15:28 ERROR unit.tempo/0.juju-log resource patcher failed with message 'Kubernetes resources patch failed: Apply failed with 1 conflict: conflict with "juju" using apps/v1: .spec.template.spec.contain
ers[name="charm"].resources.requests.memory'
unit-tempo-0: 10:15:28 INFO unit.tempo/0.juju-log HTTP Request: GET https://10.152.183.1/apis/apps/v1/namespaces/test-tempo/statefulsets/tempo "HTTP/1.1 200 OK"
unit-tempo-0: 10:15:28 INFO unit.tempo/0.juju-log HTTP Request: PATCH https://10.152.183.1/apis/apps/v1/namespaces/test-tempo/statefulsets/tempo?fieldManager=ResourcePatcher&dryRun=All "HTTP/1.1 409 Conflict"
unit-tempo-0: 10:15:29 ERROR unit.tempo/0.juju-log resource patcher failed with message 'Kubernetes resources patch failed: Apply failed with 1 conflict: conflict with "juju" using apps/v1: .spec.template.spec.contain
ers[name="charm"].resources.requests.memory'
```

### Additional context

Even if juju were to fix this, 3.6.9 would remain broken (and it'd take time). So we may need a release on our side with force=True so the charms aren't broken for now.

Our only stable track is 1/
We need to guarantee successful in-track upgrades. 

https://github.com/juju/juju/issues/20446

PietroPasotti:
Agreement on the way forward:
[...]

PR #129, opened 2025-08-21
feat: raise on patching charm container

PietroPasotti:
Adds a warning to mitigate https://github.com/canonical/observability-libs/issues/128#issuecomment-3209698155
Whoever tries to use the KubernetesComputeResourcePatch on the `charm` container will now get an error unless they pass an explicit override.

### DIFF
--- a/lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
+++ b/lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
@@ -616,7 +616,7 @@ class KubernetesComputeResourcesPatch(Object):
     def _on_config_changed(self, _):
         self._patch()
 
-    def _patch(self) -> None:
+    def _patch(self) -> None: # noqa: C901
         """Patch the Kubernetes resources created by Juju to limit cpu or mem.
 
         This method will keep on retrying to patch the kubernetes resource for a default duration of 20 seconds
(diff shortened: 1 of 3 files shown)


lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -569,6 +569,7 @@ def __init__(
         *,
         resource_reqs_func: Callable[[], ResourceRequirements],
         refresh_event: Optional[Union[BoundEvent, List[BoundEvent]]] = None,
+        allow_patching_charm_container: bool = False
```
sed-i, line 572: Given the position of the juju team, the lib should not act on "charm" and rather fail.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
```diff
@@ -582,8 +583,17 @@ def __init__(
               only raise ValueError.
             refresh_event: an optional bound event or list of bound events which
                 will be observed to re-apply the patch.
+            allow_patching_charm_container: Whether to allow patching the "charm" container.
         """
         super().__init__(charm, "{}_{}".format(self.__class__.__name__, container_name))
+
+        if container_name == "charm" and not allow_patching_charm_container:
```
sed-i, line 590: To act on `allow_patching_charm_container` we'd need a `force`?
In any case, I think we shouldn't act on `charm`.

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
PietroPasotti: but what if the user has an `assumes: juju<3.6.8`?

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
PietroPasotti: I'm not saying we should do force in any way, just let the user try if they know better (e.g they know they are on a juju version where this isn't a problem)

sed-i (approved):
Thanks!

